PR #5, opened 2025-09-10
WCP integration + cone tracing

toddstrader:
Draft for now until we get things working cleanly across the various platforms.  But it is functional and locally tested on linux.

### DIFF
--- a/include/ast/ReferenceIndexer.h
+++ b/include/ast/ReferenceIndexer.h
@@ -1,0 +1,79 @@
+//------------------------------------------------------------------------------
+//! @file ReferenceIndexer.h
+//! @brief Variable reference index
+//
+// SPDX-FileCopyrightText: Hudson River Trading
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#pragma once
+
+#include <type_traits>
+#include <unordered_map>
+
+#include "slang/ast/ASTVisitor.h"
+#include "slang/ast/Expression.h"
+#include "slang/ast/Symbol.h"
+#include "slang/ast/expressions/MiscExpressions.h"
+#include "slang/ast/symbols/BlockSymbols.h"
+#include "slang/ast/symbols/InstanceSymbols.h"
+#include "slang/ast/symbols/MemberSymbols.h"
+#include "slang/ast/symbols/ValueSymbol.h"
+
+struct ReferenceIndexer : public slang::ast::ASTVisitor<ReferenceIndexer, true, true> {
+private:
+    const slang::ast::Symbol* currentUse = nullptr;
+
+public:
+    template<typename T>
+        requires std::is_base_of_v<slang::ast::ValueExpressionBase, T>
+    void handle(const T& symbol) {
+        if (currentUse) {
+            const slang::ast::ValueSymbol* instantiatedSymbol = &(symbol.symbol);
+            if (const auto modport = instantiatedSymbol->as_if<slang::ast::ModportPortSymbol>()) {
+                instantiatedSymbol = modport->internalSymbol->as_if<slang::ast::ValueSymbol>();
+            }
+            if (instantiatedSymbol) {
+                symbolToUses[instantiatedSymbol].insert(currentUse);
+            }
+        }
+        visitDefault(symbol);
+    }
+
+    template<typename T>
+    // TODO -- port maps, other stuff?
+        requires std::derived_from<T, slang::ast::ContinuousAssignSymbol> ||
+                 std::derived_from<T, slang::ast::ProceduralBlockSymbol>
+    void handle(const T& symbol) {
+        SLANG_ASSERT(!currentUse);
+        currentUse = &symbol;
+        visitDefault(symbol);
+        currentUse = nullptr;
+    }
+
+    void handle(const slang::ast::InstanceSymbol& symbol) {
+        currentUse = &symbol;
+        for (const auto connection : symbol.getPortConnections()) {
+            const auto port = connection->port.as_if<slang::ast::PortSymbol>();
+            if (port) {
+                auto value = port->internalSymbol->as_if<slang::ast::ValueSymbol>();
+                if (value) {
+                    symbolToUses[value].insert(currentUse);
+                }
+                auto expr = connection->getExpression();
+                if (expr) {
+                    expr->visit(*this);
+                }
+            }
+        }
+        currentUse = nullptr;
+        visitDefault(symbol);
+    }
+
+    void reset(const slang::ast::Symbol* root) {
+        symbolToUses.clear();
+        root->visit(*this);
+    }
+
+    std::unordered_map<const slang::ast::ValueSymbol*, std::set<const slang::ast::Symbol*>>
+        symbolToUses;
+};
--- a/include/ast/SlangServerWcp.h
+++ b/include/ast/SlangServerWcp.h
@@ -1,0 +1,24 @@
+//------------------------------------------------------------------------------
+/// @file WcpClient.h
+/// @brief Interface for WCP -> slang server
+//
+// SPDX-FileCopyrightText: Hudson River Trading
+// SPDX-License-Identifier: MIT
+//------------------------------------------------------------------------------
+#pragma once
+
+#include <mutex>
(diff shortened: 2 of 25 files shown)


include/ast/WcpClient.h
```diff
@@ -0,0 +1,155 @@
...
+#include <thread>
+
+#ifdef _WIN32
+#    include <winsock2.h>
+#    include <ws2tcpip.h>
+#    pragma comment(lib, "ws2_32.lib")
+typedef int socklen_t;
+typedef SSIZE_T ssize_t;
+#    define close closesocket
+#    define MSG_DONTWAIT 0
```
AndrewNolte, line 26: Formatting seems pretty odd here? Might need to rerun precommit

tests/data/test2.sv
```diff
@@ -0,0 +1,92 @@
+interface intf();
```
AndrewNolte, line 1: Could we put these files in subfolder in data/, like data/wcp/

src/SlangServer.cpp
```diff
@@ -254,6 +272,127 @@ std::vector<hier::HierItem_t> SlangServer::getScope(const std::string& hierPath)
     return m_driver->comp->getScope(hierPath);
 }
 
+// TODO -- Underlying ServerCompilation implementation is slow for larger designs -- fix
```
AndrewNolte: InstanceVisitor*

include/lsp/SnippetString.h
```diff
@@ -195,7 +195,7 @@ class SnippetString {
         if (number.has_value()) {
             // If a specific number is provided, use it, but also make sure our
             // auto-increment counter is at least one greater for the next call.
-            _tabstop = std::max(_tabstop, number.value() + 1);
+            _tabstop = (std::max)(_tabstop, number.value() + 1);
```
AndrewNolte: ??

include/util/Converters.h
```diff
@@ -28,6 +28,9 @@ lsp::Position toPosition(const SourceLocation& loc, const SourceManager& sourceM
 
 lsp::Range toRange(const SourceRange& range, const SourceManager& sourceManager);
 
+lsp::Range toRange(const SourceLocation& loc, const SourceManager& sourceManager,
+                   const size_t length);
```
AndrewNolte, line 32: weird formatting again, don't think this will pass the precommit checks now

include/ast/SlangServerWcp.h
```diff
@@ -0,0 +1,21 @@
...
+//------------------------------------------------------------------------------
+#pragma once
+
+#include <mutex>
+#include <string>
+#include <vector>
+
+class SlangServerWcp {
+public:
+    virtual void gotoDeclaration(const std::string&) = 0;
```
AndrewNolte: Can we get some docs here and in slangserver.h? And can these just match the wcp route names? Would also be nice to link to the wcp docs, if those exist

src/ast/WcpClient.cpp
```diff
@@ -0,0 +1,319 @@
...
+            },
+            *s2cMessage);
+    }
+
+    stop();
+}
+
+void waves::WcpClient::onWaveformsLoaded(const wcp::WaveformsLoaded& waveformsLoaded) {
+    std::cerr << "WCP loaded waveform: " << waveformsLoaded.source << std::endl;
+    try {
```
AndrewNolte: We can have the try/catch in the dispatcher that calls these methods, similar to how the Lsp dispatcher was written

include/ast/ReferenceIndexer.h
```diff
@@ -0,0 +1,87 @@
...
+            if (instantiatedSymbol) {
+                symbolToUses[instantiatedSymbol].insert(currentUse);
+            }
+        }
+        visitDefault(symbol);
+    }
+
+    template<typename T>
+    // TODO -- port maps, other stuff?
+        requires is_base_of_any<T, slang::ast::ContinuousAssignSymbol,
```
AndrewNolte: I think this could be 
```
template<typename T>
    requires std::derived_from<T, slang::ast::ContinuousAssignSymbol> ||
             std::derived_from<T, slang::ast::ProceduralBlockSymbol>
```

include/lsp/SnippetString.h
toddstrader: It's a Windows thing.  However, I'm seeing a more sane solution here:
https://stackoverflow.com/questions/11544073/how-do-i-deal-with-the-max-macro-in-windows-h-colliding-with-max-in-std
Lemme try that first.

src/SlangServer.cpp
toddstrader: I meant the `ServerCompilation::getInstances()` impl.  But sure.

include/lsp/SnippetString.h
toddstrader: `NOMINMAX` is less awful.

include/ast/WcpClient.h
toddstrader: I'm with you that this looks weird.  I didn't even know you could put whitespace there.  However, it's `clang-format` who's doing it.  My editor keep putting it this way if I change it and if I disable that and run `pre-commit` then `clang-format` will put the spaces back.

tests/data/test2.sv
toddstrader: I renamed to `cone_test` and `wcp_test` respectively since one is not WCP.  I could create two dirs, but since I only have one test per right now this seemed better.

include/util/Converters.h
toddstrader: `pre-commit` does seem to want it this way.  Also, see the `ServerCompilation` constructor for another example of this formatting.  I don't personally care, but it is what the formatter wants.

include/ast/SlangServerWcp.h
toddstrader: You think we should have the same docs on both the pure virtual and the final methods?  Seems like those will not stay in sync.  I'll do one for now and we can copy if you feel strongly about it.

Also we sorta do match the WCP terminology already.  For one, WCP events are snake case, so `goto_declaration` ~= `gotoDeclaration`.  The only mismatch here is `add_drivers` vs `getDrivers` (and similarly for loads).  I'd argue that the method names are correct because they say what they do.  `SlangServerWcp::getDrivers()` gets a list of driver paths.  On the other hand, `WcpClient::onAddDrivers()` uses `getDrivers()` and is the method that actually does the adding (i.e. sends the message with the drivers back up WCP).

AndrewNolte: wrt the names I was mainly talking about waveformLoaded(). waveformLoaded() sounds like it returns a bool, but it's actually a callback= "onWaveformLoaded"

AndrewNolte: I don't see docs on either atm, as long as it's on one I think that would help. I think by default clangd grabs docs from the furthest implemented header, so SlangServer.h would be best.

toddstrader: Renamed the callbacks to `on` and moved the docs to `SlangServer.h`.

src/ast/ServerCompilation.cpp
```diff
@@ -120,4 +127,114 @@ std::vector<hier::HierItem_t> ServerCompilation::getScope(const std::string& hie
...
+        }
+        if (type->isSimpleBitVector()) {
+            return true;
+        }
+    }
+
+    return false;
+}
+
+std::optional<lsp::ShowDocumentParams> ServerCompilation::gotoDeclaration(const std::string& path) {
```
AndrewNolte: Maybe setHierarchyPath() or openHierarchyPath(). Since this should also reflect on the hierarchy view and modules view.

toddstrader: Sure, but `gotoDeclaration` does mirror the `goto_declaration` WCP command currently.  I can change it here as I do agree that your suggestions encompass more of what slang-server can do.  However, I think we'll leave WCP and Surfer alone for now.

toddstrader: Went with `getHierDocParams()` instead since the `ServerCompilation` method isn't the one doing the setting or opening.  It's just getting.

PR #7, opened 2025-09-10
hovers/gotos for lookup selectors (struct members and others)

AndrewNolte:


src/document/ShallowAnalysis.cpp
```diff
@@ -159,6 +159,53 @@ const ast::Symbol* ShallowAnalysis::handleInterfacePortHeader(const parsing::Tok
...
+        symbol->kind == ast::SymbolKind::Instance) {
+        if (auto varSym = symbol->as_if<ast::ValueSymbol>()) {
+            auto& type = varSym->getType().getCanonicalType();
+            switch (type.kind) {
+                case ast::SymbolKind::PackedStructType:
+                case ast::SymbolKind::UnpackedStructType:
+                case ast::SymbolKind::PackedUnionType:
+                case ast::SymbolKind::UnpackedUnionType:
+                case ast::SymbolKind::ClassType:
+                case ast::SymbolKind::ClassProperty:
```
toddstrader: Is `ClassProperty` not possible above?  And if it is, can the two lists be collapsed?

tests/cpp/golden/FindSymbolRef.out
```diff
@@ -1216,6 +1216,7 @@ module m7;
     int i = g2.foo();
```
toddstrader, line 1216: Do we not have `struct.sub_struct.foo` examples in here?  Would be nice to test this.

toddstrader (approved):
Modulo comments

src/document/ShallowAnalysis.cpp
AndrewNolte: refactored to be more generic for type categories

tests/cpp/golden/FindSymbolRef.out
AndrewNolte: added

PR #9, opened 2025-09-12
Check in vscode client

AndrewNolte:


clients/vscode/languages/sv/systemverilog.snippets.json
```diff
@@ -0,0 +1,146 @@
...
+    "body": "${1:data_type} ${2:queue_name}[\\$];",
+    "description": "insert queue."
+  },
+  "mailbox": {
+    "prefix": "mailbox",
+    "body": ["mailbox mbx", "${1:mbx = new();}"],
+    "description": "insert mailbox instance"
+  },
+  "associative array": {
+    "prefix": "AA",
```
toddstrader: What is `AA`?  Is that the thing you type to get autocompletion for an associative array?

AndrewNolte: Yeah- these date all the way back to the original extension. We should look at generating these from syntax.txt, they're essentially all simple syntaxes that start with a keyword.

toddstrader: Cool, yeah.  I was wondering about generation as well.

AndrewNolte: I guess they could be part of the completions route? It would still be static data, but could at least only show them for lhs completions.

PR #15, opened 2025-09-12
CI: run neovim plugin tests

toddstrader:
Going to try to run the neovim plugin test workflow in this repo.  Testing with this branch / PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,3 +64,22 @@ jobs:
       run: ctest --test-dir build/${{ matrix.preset }} --output-on-failure --no-tests=error -j8
       # Don't run tests on windows
       if: matrix.os != 'windows-latest'
+    # Making the neovim plugin the top-level dir below this point for busted action
+    # Moving the old checkout is easier than doing some Windows-portable rm -rf thing
+    - name: Move build
+      run: |
+        echo ${{ runner.temp }}/slang-build/bin >> $GITHUB_PATH
+        mv ${{ github.workspace }}/build/${{ matrix.preset }} ${{ runner.temp }}/slang-build
+        mv ${{ github.workspace }}/clients/neovim ${{ runner.temp }}/neovim-plugin
+        mv ${{ github.workspace }} ${{ runner.temp }}/slang-server-checkout
+        mv ${{ runner.temp }}/neovim-plugin ${{ github.workspace }}
+      working-directory: ${{ runner.temp }}
+    - name: Debug
+      run: |
+        pwd
+        ls
+    - name: Run Neovim plugin tests
+      # This should work on Windows as well, but there's some linker error
+      # that I don't feel like debugging via GH
+      if: runner.os != 'Windows'
+      uses: nvim-neorocks/nvim-busted-action@v1


clients/neovim/lua/slang-server/_core/config.lua
```diff
@@ -0,0 +1,58 @@
...
+   },
+   kinds = {
+      instance = { icon = "", hl = "SlangServerInstance" },
+      instancearray = { icon = "", hl = "SlangServerInstanceArray" },
+      scope = { icon = "󰅩", hl = "SlangServerScope" },
+      scopearray = { icon = "󰅩", hl = "SlangServerScopeArray" },
+      package = { icon = "📦", hl = "SlangServerPackage" },
+      port = {
+         input = { icon = "", hl = "SlangServerPortInput" },
+         output = { icon = "", hl = "SlangServerPortOutput" },
```
AndrewNolte: Did these icons get messed up? Or are they just weird on github?

clients/neovim/spec/example_spec.lua
```diff
@@ -0,0 +1,83 @@
...
+   assert(client)
+   -- wait for client to attach to this buffer
+   -- TODO -- something not right here, sometimes get:
+   -- vim/_editor.lua:0: nvim_exec2(): Vim:No client found
+   --
+   -- stack traceback:
+   --         vim/_editor.lua: in function 'cmd'
+   --         spec/example_spec.lua:60: in function <spec/example_spec.lua:57>
+   --
+   vim.wait(5000, function()
```
AndrewNolte: This seems a bit imprecise

toddstrader: The name is a bit misleading, but it's a timeout:
https://neovim.io/doc/user/lua.html#vim.wait()

clients/neovim/lua/slang-server/_core/config.lua
toddstrader: They're from NerdFont:
https://www.nerdfonts.com/
which is recommended by LazyVim:
https://www.lazyvim.org/

We should make a note of the dependency in the README.  I'll add that.

clients/neovim/spec/example_spec.lua
toddstrader: Also, this is test code for the plugin not the plugin code itself.  So in this case the test is just waiting for the LSP connection with `slang-server` to be set up.  The test will fail if it takes > 5 s.

toddstrader:
Pending #19

.github/workflows/build.yml
```diff
@@ -64,3 +64,22 @@ jobs:
...
       if: matrix.os != 'windows-latest'
+    # Making the neovim plugin the top-level dir below this point for busted action
+    # Moving the old checkout is easier than doing some Windows-portable rm -rf thing
+    - name: Move build
+      run: |
+        echo ${{ runner.temp }}/slang-build/bin >> $GITHUB_PATH
+        mv ${{ github.workspace }}/build/${{ matrix.preset }} ${{ runner.temp }}/slang-build
+        mv ${{ github.workspace }}/clients/neovim ${{ runner.temp }}/neovim-plugin
+        mv ${{ github.workspace }} ${{ runner.temp }}/slang-server-checkout
+        mv ${{ runner.temp }}/neovim-plugin ${{ github.workspace }}
```
toddstrader, line 75: This is a bit janky but it allows me to use the GH action below.  At some point I can deconstruct that and make something that will tolerate our subdir (or push a change to the existing action).  But for now as long as this is at the end of the build steps we should be fine.

clients/neovim/lua/slang-server/_core/config.lua
toddstrader: Resolved in #19

PR #49, opened 2025-09-26
have config lists concat

AndrewNolte:
Stacked PRs:
 * #54
 * #53
 * #52
 * #50
 * __->__#49

--- --- ---

### Have config lists concat
Also:
- Adds more logs for config
- Make glob timing logs per glob

commit-id:b1f77fd2

### DIFF
--- a/docs/start/config.md
+++ b/docs/start/config.md
@@ -10,9 +10,11 @@ The server uses a hierarchical configuration system, layering options in this or
 - `${workspaceFolder}/.slang/server.json` (should be on source control)
 - `${workspaceFolder}/.slang/local/server.json` (`.slang/local` should be ignored by source control)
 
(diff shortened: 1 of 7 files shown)


docs/start/config.md
```diff
@@ -10,9 +10,11 @@ The server uses a hierarchical configuration system, layering options in this or
 - `${workspaceFolder}/.slang/server.json` (should be on source control)
 - `${workspaceFolder}/.slang/local/server.json` (`.slang/local` should be ignored by source control)
 
+A set config option overrides the existing value in the previous config, except for lists, which are appended.
```
toddstrader, line 13: Is there a simple way we can provide the ability to do both (extend and override)?

AndrewNolte: Atm it makes sense for all the lists in the config. I think we can solve this later if we need to.

PR #63, opened 2025-10-10
Neovim plugin documentation

toddstrader:
Probably could use some pictures.  Gonna render it first and see how it looks.


docs/hdl/neovim.md
```diff
@@ -7,30 +7,84 @@ hide:
...
+If there is more than one instance of the signal then first a list of instances will be presented.
+After an instance is selected (or if there is only one instance) a list of drivers or loads will be presented.
 
 ## Waveform Integration (experimental)
 
-<div class="grid" markdown>
-<div class="grid-item" markdown>
+![WCP](neovim/wcp.gif)
 
 This is currently only setup between Neovim and Surfer, through the WCP (waveform control protocol), which runs on JSON-RPC over TCP.
```
AndrewNolte: I think we can lose the "This is currently only setup between Neovim and Surfer", since we're already on the neovim page now. Maybe "The surfer waveform viewer is integrated via WCP"

PR #65, opened 2025-10-13
vaporview integration

AndrewNolte:
### vaporview integration
  - In Vaporview, press 'e' on a selected signal, or use the buttons to open the corresponding filelist and go to that signal
  - In Slang's Hierarchy view, press 'w' or the buttons to open the corresponding waveform file and open in the waves
#### Button changes
  - Press `cmd+f` in the hierarchy or modules view to search for an instance in the design. Can also use the button to get to this command (magnifying glass)
  - Select buildfile button now looks like a folder
  - New 'Open buildfile' opens the current .f file. Saving this file will refresh a compilation. In the future this will also be generated for compilations set by top level, making it so that users can adjust the params and defines
  - "Copy Path" can be done now with either `cmd+c` or by right clicking on a signal
####  Backend fixes that apply to both vscode and neovim:
  - Type aliases and resolved types are shown in the hierarchy view
  - Macro defined symbols now show the proper locations of the macro usage that created them

### DIFF
--- a/clients/vscode/src/sidebar/VaporviewApi.ts
+++ b/clients/vscode/src/sidebar/VaporviewApi.ts
@@ -1,0 +1,283 @@
+/**
+ * Vaporview API TypeScript Functions
+ *
+ * Async wrapper functions for the Vaporview waveform viewer VSCode extension API
+ * Based on: https://github.com/Lramseyer/vaporview/blob/main/API_DOCS.md
+ */
+
+import * as vscode from 'vscode'
+import { setTimeout } from 'timers'
+import { URLSearchParams } from 'url'
+
+/**
+ * Opens a waveform dump file with Vaporview
+ */
+export async function openFile(args: {
+  uri: vscode.Uri
+  loadAll?: boolean
+  maxSignals?: number
+}): Promise<void> {
+  await vscode.commands.executeCommand('vaporview.openFile', args)
+  // Subsequent commands may fail if executed too quickly :(
+  await new Promise((r) => setTimeout(r, 200))
+}
+
+/**
+ * Variable identifier - must specify at least one of:
+ * 1. netlistId
+ * 2. instancePath
+ * 3. scopePath AND name
+ */
+export interface VariableIdentifier {
+  uri?: vscode.Uri
+  netlistId?: number
+  instancePath?: string
+  scopePath?: string
+  name?: string
+  msb?: number
+  lsb?: number
+}
+
+/**
+ * Add a variable to the waveform viewer
+ */
+export async function addVariable(args: VariableIdentifier): Promise<void> {
+  await vscode.commands.executeCommand('waveformViewer.addVariable', args)
+}
+
+/**
+ * Remove a variable from the waveform viewer
+ */
+export async function removeVariable(args: VariableIdentifier): Promise<void> {
+  await vscode.commands.executeCommand('waveformViewer.removeVariable', args)
+}
+
+/**
+ * Reveal a variable or scope in the netlist view
+ */
+export async function revealVariableInNetlistView(args: VariableIdentifier): Promise<void> {
+  await vscode.commands.executeCommand('waveformViewer.revealVariableInNetlistView', args)
+}
+
+export type TimeUnit = 'fs' | 'ps' | 'ns' | 'us' | 'µs' | 'ms' | 's' | 'ks'
+export type MarkerType = 0 | 1 // 0: Main Marker, 1: Alt Marker
+
+/**
+ * Set the marker or alt marker to a specific time in the viewer
+ */
+export async function setMarker(args: {
+  uri?: vscode.Uri
+  time: number
+  units?: TimeUnit
+  markerType?: MarkerType
+}): Promise<void> {
+  await vscode.commands.executeCommand('waveformViewer.setMarker', args)
+}
+
+/**
+ * Get a list of open waveform viewer documents
+ */
+export async function getOpenDocuments(): Promise<{
+  documents: vscode.Uri[]
+  lastActiveDocument: vscode.Uri
+}> {
+  return await vscode.commands.executeCommand('waveformViewer.getOpenDocuments')
+}
+
+/**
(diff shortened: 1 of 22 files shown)


include/Config.h
```diff
@@ -44,6 +44,10 @@ class Config {
     rfl::Description<"Waveform viewer command ({} will be replaced with the WCP port)",
                      std::optional<std::string>>
         wcpCommand;
+    rfl::Description<"Waveform file to open given a build (name and top variables can be passed "
+                     "with {name}, {top})",
+                     std::optional<std::string>>
+        wavesFile;
```
toddstrader: How does this work?  The only other places I see `wavesFile` are on the vscode client and not in `slang-server` proper.  Is this config passed into the client?  Even still, I'm not sure how it is used.

include/SlangServer.h
```diff
@@ -116,6 +116,13 @@ class SlangServer : public lsp::LspServer<SlangServer>, public SlangServerWcp {
     // Returns the instances of a module
     std::vector<hier::QualifiedInstance> getInstancesOfModule(const std::string moduleName);
 
+    // Returns the modules defined in a file, used for the modules view
+    std::vector<std::string> getModulesInFile(const std::string path);
+
+    // Inlay hints (resolved param values, port/param types, function arg names, signal values, etc)
+    std::optional<std::vector<lsp::InlayHint>> getDocInlayHint(
+        const lsp::InlayHintParams&) override;
```
toddstrader: Are these methods related to vaporview integration?

clients/vscode/src/sidebar/WcpTypes.ts
```diff
@@ -0,0 +1,81 @@
+//------------------------------------------------------------------------------
+/// @file WcpTypes.ts
```
toddstrader: Are these types necessary for vaporview integration?

clients/vscode/src/sidebar/VaporviewApi.ts
```diff
@@ -0,0 +1,267 @@
...
+  name?: string
+  msb?: number
+  lsb?: number
+}
+
+/**
+ * Add a variable to the waveform viewer
+ */
+export async function addVariable(args: VariableIdentifier): Promise<void> {
+  await vscode.commands.executeCommand('waveformViewer.addVariable', args)
```
toddstrader, line 45: Not sure if `waveformViewer.addVariable` is a vaporview method or where it is coming from, but I'd caution against having to know what is a variable vs scope or container.  Certainly there are answers to these things in the LRM, but waveform viewers don't all see things the same way and you might not want them to for ergonomic reasons.  E.g. `some_array[5]` may have a `some_array` container which has a `[5]` container.  Or is might just be a `some_array[5]` container.  This is the very thing I ran into in Surfer which prompted the WCP change from add_variable/scope to add_item.  And since they're both wellen based the story is likely similar.

clients/vscode/src/config.schema.json
```diff
@@ -0,0 +1,82 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$ref": "#/definitions/Config",
```
toddstrader: Would be nice to connect the dots between this file and `Config.h`.

include/Config.h
```diff
@@ -44,6 +44,10 @@ class Config {
     rfl::Description<"Waveform viewer command ({} will be replaced with the WCP port)",
                      std::optional<std::string>>
         wcpCommand;
+    rfl::Description<"Waveform file to open given a build (name and top variables can be passed "
+                     "with {name}, {top})",
```
toddstrader: Is there a way to open an arbitrary waves file?

include/Config.h
AndrewNolte: Yeah, the whole config is synced to the client. It's similar to `buildPattern`, and I'm thinking about naming it to match. Basically when requesting to open a signal in the waves, it can try to guess the waves file name, or at least a glob of possibilities. I was also thinking about adding `{path}` to allow for deeply nested .f and waves files.

include/SlangServer.h
AndrewNolte: Yeah I was thinking about adding these here to show values inline, but I think I'll do it it on a separate pr since this one is already pretty big.

clients/vscode/src/sidebar/WcpTypes.ts
AndrewNolte: No, I was going to use these for the next commits in the stack

clients/vscode/src/sidebar/VaporviewApi.ts
AndrewNolte: Yeah this is just how the api is right now; we might need to do a similar thing with vaporview. Surfer pr for reference: https://gitlab.com/surfer-project/surfer/-/merge_requests/713

clients/vscode/src/config.schema.json
AndrewNolte: I think you mean have config.schema.json reference Config.h? That'd be a reflect-cpp feature. This is generated via   this script in `package.json` `"genconfig": "../../build/bin/slang-server --config-schema > src/config.schema.json && npx json2ts -i src/config.schema.json -o src/config.gen.ts"`. I could probably add a @generated tag for this in the slang-server binary, or even make it part of the npm build.

include/Config.h
AndrewNolte: Yeah you can just open the correct file in vscode if it picks the wrong one.

clients/vscode/src/config.schema.json
toddstrader: Ah, yeah.  Building on the fly is clearly the winner but if you can at least get a comment in there mentioning that it's generated and what command was run that'd be great.

include/Config.h
toddstrader: Oh yup.  I'm still thinking WCP.  But yeah, your in-vscode vaporview and vscode proper will always be intrinsically tied.

PR #83, opened 2025-10-24
adjust macro goto tests to use index

AndrewNolte:
Stacked PRs:
 * #95
 * #94
 * #92
 * #91
 * #90
 * #89
 * #84
 * __->__#83

--- --- ---

### adjust macro goto tests to use index

commit-id:f16ba12f

### DIFF
--- a/tests/cpp/golden/FindSymbolRefMacro.out
+++ b/tests/cpp/golden/FindSymbolRefMacro.out
@@ -15,6 +15,16 @@
 
 
 
+// Use macros from common_macros.svh
+
+`define BUS_WIDTH `WIDTH
+
+`define MEM_SIZE `DEPTH
+
+`define ADDR_BITS `ADDR_WIDTH
+
+
+
 // Nested macro definitions
 
 `define FOO(blah) blah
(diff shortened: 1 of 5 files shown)


tests/cpp/golden/FindSymbolRefMacro.out
```diff
@@ -64,6 +74,30 @@
...
+
+
+// Macro to instantiate memory_module
+
+`define INST_MEMORY(inst_name, dwidth) \
+
+    memory_module #(.DATA_WIDTH(dwidth)) inst_name ( \
+
+        .clk(clk), \
+
```
toddstrader: What's with the extra spaces?

tests/data/macro_test/macros.svh
```diff
@@ -28,6 +33,18 @@
 
 `define INFO(msg) $display("Info: %s", msg);
 
+
+// Macro to instantiate memory_module
+`define INST_MEMORY(inst_name, dwidth) \
```
toddstrader: Could / should something use this new macro?

PR #85, opened 2025-10-24
Versioning and release CI jobs

AndrewNolte:
Stacked PRs:
 * __->__#85

--- --- ---


alex-torregrosa:
Hi! Could you consider performing the Linux builds under a container for compatibility with older glibc versions? As you may know, many EDA tools only officially support old-ish RHEL (or RHEL-compatible) distros, so binaries compiled for `ubuntu-latest` may not run on those machines.

Using the `rockylinux:8` image should do the trick, as done in the [zed-verilog-extension](https://github.com/someone13574/zed-verilog-extension/blob/main/.github/workflows/build-slang.yaml), where beta support for `slang-server` was added last week.

PR #89, opened 2025-10-27
Test for idefs attaching to next token

AndrewNolte:
Stacked PRs:
 * #95
 * #94
 * #92
 * #91
 * #90
 * __->__#89
 * #84
 * #83

--- --- ---

### Test for idefs attaching to next token

commit-id:6ea546e2


AndrewNolte:
The fix for the added test data here is #90 . This pr should be unrelated to #84

PR #106, opened 2025-11-06
installing.md: fix type of cmd for nvim

Sustrak:
Based on `lsp.lua:485`:

> validate('cmd', config.cmd, validate_cmd, 'expected function or table with executable command')


toddstrader:
Thanks for this.

PR #132, opened 2025-11-24
New indexing config for directory crawling speedup; grab all .sv, .svh, .v and vh files rather than having user specify

AndrewNolte:
Stacked PRs:
 * #148
 * #142
 * #147
 * #146
 * #145
 * #141
 * #140
 * #139
 * #138
 * __->__#132

--- --- ---

### DIFF
--- a/clients/vscode/resources/config.schema.json
+++ b/clients/vscode/resources/config.schema.json
@@ -11,14 +11,21 @@
         },
         "indexGlobs": {
           "type": "array",
-          "description": "Globs of what to index. By default will index all sv and svh files in the workspace.",
+          "description": "Deprecated: use 'index' instead. Globs of what to index. By default will index all sv and svh files in the workspace.",
           "items": {
             "type": "string"
           }
         },
+        "index": {
+          "type": "array",
+          "description": "Index configurations; by default indexes all .sv, .svh, .v, and .vh files in the workspace.",
+          "items": {
+            "$ref": "#/$defs/Config__IndexConfig"
+          }
+        },
         "excludeDirs": {
           "type": "array",
-          "description": "Directories to exclude",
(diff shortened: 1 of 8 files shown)


clients/vscode/resources/config.schema.json
```diff
@@ -85,6 +92,31 @@
...
+        "excludeDirs": {
+          "anyOf": [
+            {
+              "type": "array",
+              "items": {
+                "type": "string"
+              }
+            },
+            {
+              "type": "null"
```
toddstrader, line 116: Is `null` the default / can you specify one of these objects without providing an `excludeDirs`?

src/Indexer.cpp
```diff
@@ -327,6 +325,79 @@ std::vector<fs::path> Indexer::getFilesForMacro(std::string_view name) const {
...
+        // No index configs - index entire workspace
+        if (workspaceFolder.has_value()) {
+            collectFilesFromDirectory(fs::path(*workspaceFolder), {}, pathsToIndex);
+        }
+    }
+    else {
+        for (const auto& cfg : indexConfigs) {
+            for (const auto& dir : cfg.dirs) {
+                fs::path fullDirPath;
+                if (fs::path(dir).is_absolute()) {
```
toddstrader, line 382: What if someone excludes an absolute dir?

clients/vscode/resources/config.schema.json
AndrewNolte: Yeah, null indicates that you can leave excludeDirs out

src/Indexer.cpp
AndrewNolte: I guess I should specify that excludeDirs are just single directory names, not longer paths, which would also mean abs paths are invalid for that config

PR #152, opened 2025-12-04
Completion: Recommend keywords on the LHS of expressions

mlyoung101:
This PR makes it so that the completion system will recommend certain keywords on the LHS of expressions, currently: `logic, assign, wire, reg` and snippets for `always_comb, always_latch` and `always_ff`.

I also updated the unit tests too.

This is more or less a direct-ish port of code I wrote a few years back for my own SystemVerilog language server: https://github.com/mattyoung101/slingshot/blob/legacy/src/main/kotlin/slingshot/completion/CompletionGenerator.kt#L66

My plan in the future is to also recommend user-defined types, like the ones you create in a `typedef enum { ... } Type_t;`

### DIFF
--- a/clients/vscode/languages/sv/systemverilog.snippets.json
+++ b/clients/vscode/languages/sv/systemverilog.snippets.json
@@ -1,31 +1,4 @@
 {
-  "always_ff block": {
-    "prefix": "always_ff",
-    "body": [
-      "always_ff @( ${1:clock} ) begin : ${2:blockName}",
-      "\t$0",
-      "end"
-    ],
-    "description": "Insert an always_ff block"
-  },
-  "always_comb block": {
-    "prefix": "always_comb",
-    "body": [
-      "always_comb begin : ${1:blockName}",
-      "\t$0",
-      "end"
-    ],
-    "description": "Insert an always_comb block"
-  },
-  "always_latch block": {
-    "prefix": "always_latch",
-    "body": [
-      "always_latch begin : ${1:blockName}",
-      "\t$0",
-      "end"
-    ],
-    "description": "Insert an always_latch block"
-  },
   "bit": {
     "prefix": "bit",
     "body": "bit ${1:bit_name} = ${2:value};"
@@ -45,10 +18,6 @@
   "byte": {
     "prefix": "byte",
     "body": "byte ${1:byte:_name} = ${2:value};"
-  },
-  "logic": {
-    "prefix": "logic",
-    "body": "logic ${1:logic_name} = ${2:value};"
   },
   "packed": {
     "prefix": "packed",
--- a/clients/vscode/languages/verilog/verilog.snippets.json
+++ b/clients/vscode/languages/verilog/verilog.snippets.json
@@ -36,42 +36,6 @@
     ],
     "description": "set module, mod i0 (.*);"
   },
-  "always": {
-    "prefix": [
-      "al",
-      "always"
-    ],
-    "body": [
-      "always @($1) begin",
-      "\t$2",
-      "end"
-    ],
-    "description": "Insert an always block"
-  },
-  "alwaysposclk": {
-    "prefix": [
-      "alclk",
-      "alwaysposclk"
-    ],
-    "body": [
-      "always @(posedge clk $1) begin",
-      "\t$2",
-      "end"
-    ],
-    "description": "always @(posedge clk)"
-  },
-  "alwaysnegclk": {
-    "prefix": [
-      "alnegclk",
-      "alwaysnegclk"
-    ],
-    "body": [
-      "always @(negedge clk $1) begin",
-      "\t$2",
-      "end"
-    ],
-    "description": "always @(negedge clk)"
-  },
   "begin/end": {
     "prefix": "begin",
     "body": [
@@ -125,13 +89,6 @@
     ],
     "description": "casez () ... endcase"
   },
-  "reg": {
-    "prefix": "reg",
-    "body": [
-      "reg $1;"
-    ],
-    "description": "reg reg_name;"
-  },
   "regarray": {
     "prefix": [
       "regarray",
@@ -151,13 +108,6 @@
       "reg [$1:$2] $3 [$4:$5];"
     ],
     "description": "reg [N:0] reg_name [0:M];"
-  },
-  "wire": {
-    "prefix": "wire",
-    "body": [
-      "wire $1;"
-    ],
-    "description": "wire wire_name;"
   },
   "wirearray": {
     "prefix": [


tests/cpp/golden/WildcardImportCompletion.json
```diff
@@ -249,6 +249,32 @@
...
+        "kind": "Snippet",
+        "filterText": "always_ff",
+        "insertText": "always_ff @($0) begin\n\nend",
+        "insertTextFormat": "Snippet"
+      },
+      {
+        "label": "always_latch",
+        "kind": "Snippet",
+        "filterText": "always_latch",
+        "insertText": "always_latch begin\n$0\nend",
```
AndrewNolte: Should there be a tab here? I believe the client will correct \t to whatever the tab format is.

AndrewNolte:
Looks good!
- User defined types should already come up in lhs completions.
- Ideally something like the script in `external/slang/scripts/syntax_gen.py` could generate these for constructs that start with a keyword
- Are you using vscode? There's already some snippets in `clients/vscode/languages/sv/systemverilog.snippets.json`, in which case we should delete the duplicates there

tests/cpp/golden/WildcardImportCompletion.json
mlyoung101: So you're saying basically `always_latch begin\n\t\$0\nend`? It's probably not a bad idea. I didn't know about that client trick, that's cool. I'm happy to add this, I just didn't see it locally because I have a plugin that strips whitespace from the file on save.

mattyoung101:
Cool! Thank you.

For the user defined types, yes, after I submitted this PR I did actually see that they come up. Although, for some reason they're not showing for `typedef enums`. 

I like the idea about using Slang's syntax gen script, I'm just not sure how we'd plumb that through the CMake build, or if it's a thing we'd generate once off. The language doesn't change that often after all.

I'm using Neovim myself, so I don't get those VSCode snippets. It may be better to provide the snippets at the LSP level to have better cross-editor compatibility.

mattyoung101:
Having a look at it now, but for whatever reason the server keeps crashing on me when I edit test files. I think there's been a regression, but I don't think this branch introduced it as it happens on the tip of `main` too. Here's the backtrace:

```
* thread #1, name = 'slang-server', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
    frame #0: 0x000057247c58316f slang-server`server::InlayHintCollector::handle(slang::syntax::HierarchyInstantiationSyntax const&) + 1343
slang-server`server::InlayHintCollector::handle:
->  0x57247c58316f <+1343>: movq   (%rax,%rcx,8), %rax
    0x57247c583173 <+1347>: movq   0x8(%rax), %rcx
    0x57247c583177 <+1351>: movq   0x10(%rax), %rax
    0x57247c58317b <+1355>: movq   %rax, 0x160(%rsp)
(lldb) bt
* thread #1, name = 'slang-server', stop reason = signal SIGSEGV: address not mapped to object (fault address: 0x0)
  * frame #0: 0x000057247c58316f slang-server`server::InlayHintCollector::handle(slang::syntax::HierarchyInstantiationSyntax const&) + 1343
    frame #1: 0x000057247c585fb1 slang-server`server::InlayHintCollector::collectHints() + 417
    frame #2: 0x000057247c571692 slang-server`server::ShallowAnalysis::getInlayHints(lsp::Range, Config::InlayHints const&) + 34
    frame #3: 0x000057247c5efd95 slang-server`server::SlangServer::getDocInlayHint(lsp::InlayHintParams const&) + 85
    frame #4: 0x000057247c6adfcf slang-server`void lsp::JsonRpcServer<server::SlangServer>::registerMethod<lsp::InlayHintParams, std::optional<std::vector<lsp::InlayHint, std::allocator<lsp::InlayHint>>>, &server::SlangServer::getDocInlayHint(lsp::InlayHintParams const&)>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&)::'lambda'(std::optional<rfl::Generic>)::operator()(std::optional<rfl::Generic>) const::'lambda'()::operator()() const + 79
    frame #5: 0x000057247c6adeb8 slang-server`void lsp::JsonRpcServer<server::SlangServer>::registerMethod<lsp::InlayHintParams, std::optional<std::vector<lsp::InlayHint, std::allocator<lsp::InlayHint>>>, &server::SlangServer::getDocInlayHint(lsp::InlayHintParams const&)>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&)::'lambda'(std::optional<rfl::Generic>)::operator()(std::optional<rfl::Generic>) const + 40
    frame #6: 0x000057247c6ade20 slang-server`std::_Function_handler<rfl::Generic (rfl::Generic), void lsp::JsonRpcServer<server::SlangServer>::registerMethod<lsp::InlayHintParams, std::optional<std::vector<lsp::InlayHint, std::allocator<lsp::InlayHint>>>, &server::SlangServer::getDocInlayHint(lsp::InlayHintParams const&)>(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&)::'lambda'(std::optional<rfl::Generic>)>::_M_invoke(std::_Any_data const&, rfl::Generic&&) + 48
    frame #7: 0x000057247c6f712e slang-server`lsp::JsonRpcServer<server::SlangServer>::processMessage(lsp::RpcRequest) + 606
    frame #8: 0x000057247c6f183c slang-server`lsp::JsonRpcServer<server::SlangServer>::handleMessage(lsp::RpcRequest) + 76
    frame #9: 0x000057247c6e5e73 slang-server`lsp::JsonRpcServer<server::SlangServer>::run() + 1251
    frame #10: 0x000057247c6e57f8 slang-server`main + 920
    frame #11: 0x000076b36722a578 libc.so.6`__libc_start_call_main(main=(slang-server`main), argc=1, argv=0x00007fff58f2c8f8) at libc_start_call_main.h:58:16
    frame #12: 0x000076b36722a63b libc.so.6`__libc_start_main_impl(main=(slang-server`main), argc=1, argv=0x00007fff58f2c8f8, init=<unavailable>, fini=<unavailable>, rtld_fini=<unavailable>, stack_end=0x00007fff58f2c8e8) at libc-start.c:360:3
    frame #13: 0x000057247c549ba5 slang-server`_start + 37
(lldb)
```

mattyoung101:
For the other ones though:

Now that I look at `syntax_gen.py`, I'm somewhat confused about how we'd integrate it on our side. It seems to be an internal Slang script, so it generates Slang classes for the syntax tree. Is it definitely not better to just have a static array of strings that indexes all the keywords? Or maybe there's some programmatic way to iterate over all the keyword tokens once it's been generated?

I didn't see `--update`, but I did eventually find the global variable that I can set to update the tests. It was bit confusing since they get written to the build dir rather than the source dir. Do you have a particular regression testing library in mind? I would figure Catch2 could do it, but yeah some docs might be good.

src/completions/Completions.cpp
```diff
@@ -552,4 +560,41 @@ void getMemberCompletions(std::vector<lsp::CompletionItem>& results, const slang
...
+            .documentation = std::nullopt,
+            .filterText = kw,
+        });
+    }
+
+    // on the LHS, we can also provide one of the always_XX routines
+    results.emplace_back(lsp::CompletionItem{
+        .label = "always_ff",
+        .kind = lsp::CompletionItemKind::Snippet,
+        .documentation = std::nullopt,
```
AndrewNolte, line 577: This will default to nullopt, although I understand this is done in a couple of other places

mlyoung101: Yeah, I'm defaulting it to nullopt here because I figured documentation wasn't really relevant for keywords.

mattyoung101:
Yep, done that now. Let me know if I've removed the right blocks, I might have been a bit too aggressive, but I think it should be ok.

mattyoung101:
Just on the above, FYI I removed the `always_ff @({posedge/negedge})` snippet because I intend to add completion support for sensitivity lists including `posedge/negedge` in a future PR :)

src/completions/Completions.cpp
AndrewNolte: It's basically like a preview for what you insert, along with any docs. 'detail' is a nice field to have filled in too. For these it would just be "always block", or whatever the syntax kind is that is being filled in.

src/completions/Completions.cpp
```diff
@@ -552,4 +560,41 @@ void getMemberCompletions(std::vector<lsp::CompletionItem>& results, const slang
...
+        .filterText = "always_ff",
+        .insertText = "always_ff @($0) begin\n\t\nend",
+        .insertTextFormat = lsp::InsertTextFormat::Snippet,
+    });
+    results.emplace_back(lsp::CompletionItem{
+        .label = "always_comb",
+        .kind = lsp::CompletionItemKind::Snippet,
+        .documentation = std::nullopt,
+        .filterText = "always_comb",
+        .insertText = "always_comb begin\n\t$0\nend",
```
AndrewNolte, line 587: You can also label these tabstops, see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#snippet_syntax

PR #157, opened 2025-12-09
Windows Compatibility

evanwporter:
Adds compatibility with windows to `slang-server`. In particular it:
- Implement a new URI class that is compatible on windows and linux
- Adds windows-only fixes for handling streams.
- Fixes every test so it passes on windows
- Ensures it works on VSCode and Windows (neovim untested)

### DIFF
--- a/tests/cpp/utils/GoldenTest.h
+++ b/tests/cpp/utils/GoldenTest.h
@@ -145,6 +145,8 @@ public:
         std::string cwd = std::filesystem::current_path().generic_string();
 
 #ifdef _WIN32
+        // Even though `fs::current_path()` is absolute, we still check to
+        // make sure its not a UNC path (ie: `\\server\share\path\to\somewhere`)
         if (cwd.size() >= 2 && cwd[1] == ':' && std::isalpha(cwd[0]))
             cwd[0] = static_cast<char>(std::toupper(cwd[0]));
         std::string file_prefix = "file:///" + cwd + "/";


evanwporter:
Regarding URI, the underlying URI is decoded in two places which is very redundant, but the test make extensive use of the `.str()` method which returns `_underlying` directly. Ideally though, we route everything through `getPath` or `getURI`. `getURI` should return the actual URI meaning that there may be stuff like `%3A` in place of semicolons, so the test cases should be modified to expect this.

evanwporter:
@AndrewNolte I've been messing around a little more, trying to get the VScode client to work. All though (with the PR) the C++ unit tests pass, for whatever the vscode client doesn't work. Are there any tests that exist already that could allow me to test the vscode-slang server interaction?

Also FYI, the debugger reveals the program is stuck at this line: https://github.com/hudson-trading/slang-server/blob/b0542b539ebf69dcdad3f5bd1e0590f51ec99f01/include/lsp/JsonRpc.h#L87C1-L88C1

evanwporter:
Now everything except the macro test should work...

AndrewNolte:
> the github action usage for this month is going to be crazy XD
 
Free for OSS! Just slow

evanwporter:
Yeah me too. But Linux is the same and so is clangd, so that might be a
vscode issue

.github/workflows/build.yml
```diff
@@ -85,7 +85,7 @@ jobs:
     - name: Run ctest
       run: ctest --test-dir build/${{ matrix.preset }} --output-on-failure --no-tests=error -j8
       # Don't run tests on windows
-      if: runner.os != 'Windows'
+      # if: runner.os != 'Windows'
```
AndrewNolte: Can we just delete it now?

.github/workflows/build.yml
```diff
@@ -85,7 +85,7 @@ jobs:
...
     - name: Set up uv
       uses: astral-sh/setup-uv@v6
     - name: Run pytest (not Windows)
```
AndrewNolte: Did the pytests not work on windows?

evanwporter:
Yeah

src/lsp/URI.cpp
```diff
@@ -0,0 +1,198 @@
...
+    R"(^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?$)"};
+
+URI::URI(const std::string& input) {
+    auto m = ctre::match<uriPattern>(input);
+    if (!m) {
+        return; // empty URI
+    }
+
+    scheme = m.get<2>().to_string();
+    authority = decode(m.get<4>().to_string());
```
AndrewNolte: Can we decode the whole thing at once, then have the components be string_views into the underlying? Should simplify ==, and be more memory efficient.

src/lsp/URI.cpp
```diff
@@ -0,0 +1,198 @@
...
+    path = decode(m.get<5>().to_string());
+
+    // Normalize Windows drive letter to uppercase
+    if (path.size() >= 3 && path[0] == '/' && std::isalpha(path[1]) && path[2] == ':') {
```
AndrewNolte: could this be ifdef'd for windows?

.github/workflows/build.yml
evanwporter: they work now

AndrewNolte:
I've been meaning to fix the \`include file invalidation as well, since the buffer ids for each time a file is included is actually different. The fix for that should be doing a similar check to what you added, but in `ShallowAnalysis::hasValidBuffers`. Please also add a directed test for this if you need to fix it here! (edit `included file, goto def on a macro from that file). But if this can pass CI without that, I'd prefer for that to be a separate PR.

.github/workflows/build.yml
```diff
@@ -59,132 +59,136 @@ jobs:
...
-        mv ${{ runner.temp }}/neovim-plugin ${{ github.workspace }}
-      working-directory: ${{ runner.temp }}
-    - name: Run Neovim plugin tests
-      # This should work on Windows as well, but there's some linker error
-      # that I don't feel like debugging via GH
-      # Skip for container builds (Rocky Linux) as paths don't work the same way
-      if: runner.os != 'Windows' && matrix.container == ''
-      uses: nvim-neorocks/nvim-busted-action@v1
+      - name: Install dependencies (Rocky Linux 8)
+        if: matrix.container == 'rockylinux:8'
```
AndrewNolte, line 63: Why did the indent change so much here?

AndrewNolte: Unfortunately github isn't great at showing that a diff is just tabs

PR #158, opened 2025-12-12
Generate and use the compilation database

evanwporter:


.vscode/settings.json
```diff
@@ -23,5 +23,9 @@
   "[typescript]": {
     "editor.formatOnSave": true,
     "editor.defaultFormatter": "esbenp.prettier-vscode"
-  }
+  },
+  "[cpp]": {
+    "editor.defaultFormatter": "llvm-vs-code-extensions.vscode-clangd",
+    "editor.formatOnSave": true,
```
AndrewNolte: `editor.formatOnSave` should be a user setting, not a repo setting

CMakeLists.txt
```diff
@@ -23,6 +23,10 @@ if(CMAKE_SOURCE_DIR STREQUAL CMAKE_BINARY_DIR)
                 "'CMakeCache.txt' and 'CMakeFiles/' before building again.")
 endif()
 
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
AndrewNolte, line 27: We want 'debug' to remain case insensitive.
```
set(CMAKE_EXPORT_COMPILE_COMMANDS
    # This expression evaluates to 'ON' if the current config is 'Debug', 
    # and 'OFF' otherwise. The value is set directly.
    # The $<CONFIG:...> expression is inherently case-insensitive for standard configs.
    "$<IF:$<CONFIG:Debug>,ON,OFF>"
    CACHE BOOL "Enable or disable export of compile commands for the Debug configuration."
)
```

.vscode/settings.json
evanwporter: ah okay, I though this might be okay since right above is this

```
  "[typescript]": {
    "editor.formatOnSave": true,
    "editor.defaultFormatter": "esbenp.prettier-vscode"
  }
  ```

CMakeLists.txt
evanwporter: I think the solution is just to always generate a comp db. Since msbuild/visualstudio doesn't have an explicit build type, doing this could mess it up.

.vscode/settings.json
AndrewNolte: Good point, removing that in this pr- https://github.com/hudson-trading/slang-server/pull/168

PR #165, opened 2025-12-15
change formatting of hover

evanwporter:
* move comments out of the code block, and remove the comment markers from them
* this is so I can do cool stuff like adding markdown to my doc comments, also this is how clangd does it (see below)
   * there's not a standard way to compile doxygen from a system verilog file, but it might cool to provide doxygen support in the future. I know this is something that clangd has

image

* also edited how macros work to keep the indentation; see below (this is after my changes)
* this is mostly a nitpick thing

image

If a slang formatter ever comes to be then the macro should just be formatted with the formatting library. This is how clangd does it; all hover code blocks are formatted with clang-format before being displayed.

### DIFF
--- a/tests/cpp/CMakeLists.txt
+++ b/tests/cpp/CMakeLists.txt
@@ -22,9 +22,6 @@ target_include_directories(
 
 add_test(NAME server_unittests COMMAND server_unittests)
 
-set_tests_properties(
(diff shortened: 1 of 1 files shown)


evanwporter:
Closed in favor of #190

PR #169, opened 2025-12-17
[VSCode] modules should not expand when a module is clicked in the hierarchy

evanwporter:
small little nitpick; see following:

Click a module here:

image

And the module pops up:

image

This PR makes it so that it only selects the corresponding item in modules, if its visible (ie: not compressed). However perhaps the popping up is the intended behavior, in which case I can just keep this as a patch on my own fork.


AndrewNolte:
Thanks for the patch, I've been meaning to fix this!

PR #176, opened 2025-12-20
[VSCode] Add installer to download latest release

evanwporter:
* combined `getValueAsync` and `checkPathNotify` into `findSlangServer`
* moved platform specific logic into a seperate file
* remove updating of the used `slang-server` when `settings.json/slang.path` changes
* added basic installer tests

The UI flow for the installer is:

```
(1) Extension starts
(2) `slang-server` not found in `settings.json`
     - Else Goto end
(3) `slang-server` not found in PATH
    - Else Goto end
(4) `slang-server` in installation dir?
    - No? Goto (7)
(5) Check if the installed one is the latest.
    - Yes? Goto end
(6) Offer to update binary
    - Yes? Goto (8)
(7) Prompt `slang-server required but not found would you like to install it?
    - Dismiss? Goto end
(8) Install slang-server
    - Popup displaying progress of download
    - Popup displaying that the download finished and where it was installed.
(9) Profit ... or not.
```


clients/vscode/tsconfig.json
AndrewNolte: Can you add "DOM" here? It should make the CI issues go away.

AndrewNolte: like the warnings in the PR

clients/vscode/test/unit/assets/release.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "v0.2.1",
+  "tag_name": "v0.2.1",
+  "assets": [
+    {
+      "name": "slang-server-linux-x64-gcc.tar.gz",
+      "browser_download_url": "http://127.0.0.1:9999/slang-server-linux-x64-gcc.tar.gz"
```
AndrewNolte, line 7: Can these tar.gz be small and created on the fly? I don't think it's a great practice to have these in git, for both security reasons and git bloat.

evanwporter:
Yeh good idea. I’ll add a zipping lib to the dev deps and zip a text file
or something and use that to test.

evanwporter:
Ya I’ll do this soon

.github/workflows/vscode-ci.yml
```diff
@@ -15,22 +15,30 @@ jobs:
...
+
       - name: Run Lint
         working-directory: clients/vscode
         run: pnpm run lint
+
+      - name: Run Tests
+        working-directory: clients/vscode
+        run: pnpm run test
+
       # TODO: Add proper tests
```
AndrewNolte: Let's remove this todo, or move it to the tests file to denote that it's for tests related to the hierarchy commands.

evanwporter:
Forgot to remove them mb. The tests automatically generate compressed files
now, it just needs the ‘assets/release.json’

clients/vscode/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "target": "ES2020",
     "outDir": "out",
     "lib": [
-      "ES2020"
+      "ES2020",
+      "dom"
```
AndrewNolte: I think this needs to be capitalized?

clients/vscode/src/installer/prepare.ts
```diff
@@ -0,0 +1,41 @@
...
+import { getPlatform } from '../lib/platform'
+
+const BIN_NAME = getPlatform() === 'windows' ? 'slang-server.exe' : 'slang-server'
+
+export async function prepareSlangServer(ui: InstallerUI): Promise<string> {
+  const release = await latestRelease()
+  const installRoot = path.join(ui.storagePath, 'install', release.tag_name)
+
+  // Check if already installed
+  const bin = await findExistingBinary(installRoot)
```
AndrewNolte: I got an ENOENT exception around here- I think you need to `mkdir -p` the `installRoot`.

clients/vscode/src/installer/ui.ts
```diff
@@ -0,0 +1,59 @@
+import * as vscode from 'vscode'
+
+export class InstallerUI {
+  constructor(private readonly context: vscode.ExtensionContext) {}
+
+  get storagePath(): string {
+    return this.context.globalStorageUri!.fsPath
```
AndrewNolte: I think our vscode node libraries should match? On mine this is always defined, so no need for the `!`.

clients/vscode/src/installer/ui.ts
```diff
@@ -0,0 +1,59 @@
...
+            last = fraction
+          }
+        }, abort)
+      }
+    )
+  }
+
+  async promptInstall(version: string): Promise<boolean> {
+    const msg =
+      `slang-server (${version}) is required but was not found.\n` +
```
AndrewNolte: The version required is really major.minor.*, and I think this will just confuse people, so I think we should take out the version from this message.

clients/vscode/src/extension.ts
```diff
@@ -147,27 +145,38 @@ export class SlangExtension extends ActivityBarComponent {
 
     // Check for environment variable set in launch.json; set when debugging in vscode
     let slangServerPath = process.env.SLANG_SERVER_PATH
+
+    // If not set, use configured path in settings.json
     if (!slangServerPath) {
-      slangServerPath = await this.path.getValueAsync()
+      slangServerPath = await this.path.findSlangServer()
       this.logger.info(`Using slang-server at ${slangServerPath}`)
```
AndrewNolte: This branch just logs `Using slang-server at ` before prompting to install, let's clean this up by maybe extracting this logic into a function that returns true if a slang-server was found. We could also put the `Starting language server` log after we've actually found a slang-server.

clients/vscode/src/installer/ui.ts
evanwporter: I invited you to join as a collaborator for my fork, that way you can make edits directly instead of having to add a comment every time (for this pr and future prs)

evanwporter: but I'll work on this later this week--maybe tomorrow if I have time.

clients/vscode/src/extension.ts
evanwporter: So the ui flow should be:

```
(1) Extension starts
(2) `slang-server` not found in `settings.json`
     - Else Goto end
(3) `slang-server` not found in PATH
    - Else Goto end
(4) `slang-server` in installation dir?
    - Yes? Goto end
(5) Prompt `slang-server required but not found would you like to install it?
    - Dismiss? Goto end
(6) Install slang-server
(7) Profit ... or not.
```

evanwporter:
Andrew, can you check whether this works on linux? You can delete the installed exe by looking in:

`.config/Code/User/globalStorage/`

PR #179, opened 2025-12-24
[VSCode] Remove unused code

evanwporter:
* Remove `which` method which is redundant with the imported `which` pkg


evanwporter:
Andrew lmk if this is not right, but as far as i can see this method is unused, but still somehow snuck into the code. I think using the npm which library is the right move here.

evanwporter:
I made this way more complicated with all my commits but its ready to go

AndrewNolte:
Yeah this is old code, thanks!

PR #189, opened 2025-12-31
Compilation diags when design is set

AndrewNolte:
Stacked PRs:
 * __->__#189

--- --- ---

### Compilation diags when design is set

- When a design is set, semantic and analysis diags come from the design rather than the shallow compilation.
- Design is updated on saves, along with those diags. Parse/Shallow diags are still published on each change.
- Removes 'unused but set' diags from unused tagging
- Bumps slang, which had some multidriven bugs at the existing commit

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,6 +280,7 @@ add_library(
   slang_server_obj_lib OBJECT
   ${CMAKE_CURRENT_BINARY_DIR}/VersionInfo.cpp
   src/ast/ServerCompilation.cpp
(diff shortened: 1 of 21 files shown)


AndrewNolte:
> @AndrewNolte What tool are you using that creates all of the stack branches?
> 
> Edit: 245 branches lol
> 
> image

This- https://github.com/modular/stack-pr. It would be nice if github natively supported stacked prs. It technically does, but it's cumbersome to do manually; it's not as simple as pushing. With this tool I just do `spr submit` and it uploads the stack.

PR #190, opened 2026-01-06
Hovers: Refactor into new file; add more info; format info like clangd

AndrewNolte:
Stacked PRs:
 * #193
 * #192
 * #191
 * __->__#190

--- --- ---

### DIFF
--- a/src/util/Markdown.cpp
+++ b/src/util/Markdown.cpp
@@ -1,0 +1,71 @@
+// SPDX-FileCopyrightText: Hudson River Trading
+// SPDX-License-Identifier: MIT
+
+#include "util/Markdown.h"
+
+#include <fmt/format.h>
+#include <sstream>
+
+namespace server::markup {
+
+// Paragraph implementation
+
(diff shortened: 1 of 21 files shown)


AndrewNolte:
No, this won't be formatting the syntax returned in the hover. `format` here means the overall markdown structure with symbol kind, various meta data, then the syntax. We're planning on making `slang-format` sometime in the first half of this year.

src/util/Markdown.cpp
```diff
@@ -0,0 +1,71 @@
...
+    return *this;
+}
+
+Paragraph& Paragraph::appendCode(std::string_view code) {
+    fmt::format_to(std::back_inserter(buffer), "`{}`", code);
+    return *this;
+}
+
+Paragraph& Paragraph::appendBold(std::string_view text) {
+    if (buffer.back() != ' ' && !buffer.empty()) {
```
evanwporter: These should be swapped. If its empty, then buffer.back() will result in ub

PR #195, opened 2026-01-09
use full word in un used diag

evanwporter:
Recently I synced up my fork with upstream/main, and I found a bunch of great features. I also learned there was a bunch of unused variables in my code. However, it illustrates that I have an unused var by only underlying the first char of the var. This pr fixes that see:

image

may not be the cleanest fix, but it works.


evanwporter:
alright, i got busy this weekend too.

PR #200, opened 2026-01-15
clangd style hovers

evanwporter:
This PR makes adjusts how hovers display comments by moving the comments section out of the code block. By doing so we are able to apply markdown to the doc comments as in the following image:

image

This is the first part in a series of PR's that standardize the hovers in `slang-server`. Future plans entail:

1) The first line of hovers (ie : Variable X in Y) should be a level 3 header (like clangd)

2) The first line should have the variable/first word be lowercased (like clangd). For `ModportPort` we should just rename it to `modport port`.

3) (Unplanned) Doxygen comment support. Clangd recently added support to display doxygen directly into the hovers. There is no standard way to compile doxygen from a system verilog file, however adding partial doxygen support to the hovers is a very useful QoL feature.

4) A setting to switch between markdown style hovers and plain style hovers (where the markdown is escaped). See [clangd](https://github.com/hudson-trading/slang-server/pull/200).


tests/cpp/golden/FindSymbolRefHdl.out
```diff
@@ -301,24 +301,24 @@ module TestModule #(
...
-               ^ Ref -> **Modport** `master` in `bus_if`  \n\\n\\n\---\n\\n\`// Modport for master\n\modport master (\n\    input clk, rst, ready,\n\    output valid, addr, data, write_enable\n\);`
-                ^ Ref -> **Modport** `master` in `bus_if`  \n\\n\\n\---\n\\n\`// Modport for master\n\modport master (\n\    input clk, rst, ready,\n\    output valid, addr, data, write_enable\n\);`
+           ^ Ref -> **Modport** `master` in `bus_if`  \n\\n\\n\---\n\\n\ Modport for master\n\\n\  \n\\n\\n\---\n\\n\`modport master (\n\    input clk, rst, ready,\n\    output valid, addr, data, write_enable\n\);`
+            ^ Ref -> **Modport** `master` in `bus_if`  \n\\n\\n\---\n\\n\ Modport for master\n\\n\  \n\\n\\n\---\n\\n\`modport master (\n\    input clk, rst, ready,\n\    output valid, addr, data, write_enable\n\);`
+             ^ Ref -> **Modport** `master` in `bus_if`  \n\\n\\n\---\n\\n\ Modport for master\n\\n\  \n\\n\\n\---\n\\n\`modport master (\n\    input clk, rst, ready,\n\    output valid, addr, data, write_enable\n\);`
```
AndrewNolte: I'm a bit confused why these newlines show up, although it doesn't appear to affect the final render

AndrewNolte: Oh I'm reading the diff wrong, nvm

evanwporter:
> The branch could also use a rebase btw

Rebase onto 165?

src/Hovers.cpp
```diff
@@ -107,8 +112,15 @@ lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
...
+    const syntax::SyntaxNode& display_node = selectDisplayNode(*info.node);
+
+    const std::string docComments = stripDocComment(formatDocComment(display_node));
+
+    if (!docComments.empty())
+        doc.addParagraph().appendText(docComments).newLine();
+
     // Add the main code block with proper formatting
-    doc.addParagraph().appendCodeBlock(formatSyntaxNode(*info.node));
+    doc.addParagraph().appendCodeBlock(formatCode(display_node));
```
evanwporter: adding a setting to leave the comments in the code block is trivial. We'd just need to modify these lines.

src/util/Formatting.cpp
```diff
@@ -171,6 +172,84 @@ std::string detailFormat(const syntax::SyntaxNode& node) {
     return res;
 }
 
+std::string stripDocComment(std::string_view input) {
```
AndrewNolte: This would be simpler and more efficient if it was operating on the trivia objects directly, rather than having to essentially reparse the comment trivia. This would look like extracting the trivia iterator out of `SyntaxPrinter::printLeadingComments` into its own function and using that. It's fine to copy/paste that code into a function here, and we can wait for a slang PR later if we want to split it there.

src/Hovers.cpp
AndrewNolte: The hovers look pretty good now, so we'll see if people ask for that when it's merged.

PR #203, opened 2026-01-16
Neovim cells interface

toddstrader:
Adds cells (vscode client calls it modules) interface to neovim client.

### DIFF
--- a/clients/neovim/lua/slang-server/navigation/cells.lua
+++ b/clients/neovim/lua/slang-server/navigation/cells.lua
@@ -191,6 +191,11 @@ end
 function M.show()
    local navigation = require("slang-server.navigation")
    local hier = require("slang-server.navigation/hierarchy")
+
+   if not hier.state.split then
+      return
+   end
+
    local split = ui.NuiSplit({
       relative = {
          type = "win",


clients/neovim/lua/slang-server/_lsp/types.lua
```diff
@@ -36,4 +36,14 @@
 
 ---@alias slang-server.lsp.Node slang-server.lsp.Item | slang-server.lsp.Var | slang-server.lsp.Scope | slang-server.lsp.FilledInstance
 
+---@class slang-server.lsp.QualifiedInstance
```
sgaulter, line 39: Does this need to inherit from `lsp.Instance`?

clients/neovim/lua/slang-server/hierarchy.lua
```diff
@@ -336,7 +486,44 @@ local function show_nodes(nodes, parent, root, remaining_path)
...
+      cell_node._uid = "__DECL__" .. cell.declName
+
+      cell_node = vim.tbl_deep_extend("error", cell_node, cell)
+
+      ---@cast cell_node slang-server.hierarchy.CellNode
+
+      local cell_nui_node = ui.NuiTree.Node(cell_node)
+      M.state.cellTree:add_node(cell_nui_node)
+      if cell.inst then
+         show_insts({ cell.inst }, cell_nui_node)
```
sgaulter: I may be misreading, but does this expand all by default? I think my preference would be to start with all collapsed.

toddstrader: Yeah, this is the way that the vscode client does it (== 1 instance: expanded, > 1 instance: collapsed).  @AndrewNolte can you elaborate on this UI choice?

Also, I'm not sure how much I care one way or the other.

AndrewNolte: The modules with only one instance tend to be important, so I think it's nice to just have those already expanded. The inst field is populated if there's only one instance.

clients/neovim/lua/slang-server/_lsp/types.lua
toddstrader: No, it's not the same:
https://github.com/hudson-trading/slang-server/blob/39ac6fcc731fc23966f82884e8a31da4633de42d/include/ast/HierarchicalView.h#L87-L91

toddstrader:
Depends on #230

clients/neovim/lua/slang-server/navigation/cells.lua
```diff
@@ -0,0 +1,241 @@
...
+         show_insts({ cell.inst }, cell_nui_node)
+      end
+   end
+
+   M.state.tree:render()
+end
+
+function M.show()
+   local navigation = require("slang-server.navigation")
+   local hier = require("slang-server.navigation/hierarchy")
```
sgaulter, line 193: It's probably worth `nil` checking that the hierarchy split exists here. I know it always *should* when cells' `show` is called, but just to avoid errors if we get into a bad state somehow.

sgaulter (approved):
Thanks for the refactor, this is a lot clearer!

PR #214, opened 2026-01-21
Hover headers

evanwporter:
Stacks onto #200.

But I separated them to make it easier to tell what changed in the golden tests (or it will be easy once 200 merges)


AndrewNolte:
Was this merged with the other one? Please try to give better PR titles btw

evanwporter:
This pr was part 2 of making the hovers look more clangd. To accomplish
that I was going to (iirc) the type info level 2 headers in the hover
markdown text.

At this point I’m so used to seeing the current hovers, that this pr isn’t
that important any more.

PR #237, opened 2026-02-01
Trigger textDocument/didSave only on open docs

alex-torregrosa:
Some editors send `didSave` events for any saved document, and not only SystemVerilog files. See zed-industries/zed#37813. The LSP spec is not 100% clear on this, although an issue in the LSP repo (microsoft/language-server-protocol#2110) seems to say this is not allowed.

This results in `slang-server` getting `didSave` notifications for any type of document, which will load and issue diagnostics.

The proposed fix is to filter the URIs in `didSave`, allowing through only the ones that correspond to documents actually opened with `didOpen`. Maybe this can create issues if non-verilog files are included into verilog files, or if the editor can send `didSave` events to non-opened files (maybe when renaming a symbol?).

### DIFF
--- a/include/ServerDriver.h
+++ b/include/ServerDriver.h
@@ -78,6 +78,9 @@ public:
 
     void onDocDidChange(const lsp::DidChangeTextDocumentParams& params);
 
+    /// @brief Checks if a document is open
+    bool isDocumentOpen(const URI& uri);
+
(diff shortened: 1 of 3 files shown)


AndrewNolte:
Yeah clients definitely shouldn't send useless traffic like this, but seems fine to double check. Is didSave really the only affected method?

PR #242, opened 2026-02-03
Indexer: update on workspace/didChangeWatchedFiles notifs, typically activated on git pulls

AndrewNolte:
Stacked PRs:
 * #245
 * #244
 * #243
 * __->__#242

--- --- ---

### DIFF
--- a/src/Indexer.cpp
+++ b/src/Indexer.cpp
@@ -94,7 +94,7 @@ void extractDataFromTree(const slang::syntax::SyntaxTree& tree, Indexer::Indexed
 
 void populateIndexForSingleFile(const fs::path& path, Indexer::IndexedPath& dest,
                                 slang::SourceManager& sourceManager, const slang::Bag& options) {
-    dest.path = path.string();
+    // Don't set dest.path here - it will be interned in indexPath
 
     // Don't expand includes - we only want symbols defined in this file
     auto tree = slang::syntax::SyntaxTree::fromFile(path.string(), sourceManager, options);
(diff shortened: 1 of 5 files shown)


src/Indexer.cpp
```diff
@@ -375,6 +346,51 @@ bool isSystemVerilogFile(const fs::path& path) {
     return ext == ".v" || ext == ".sv" || ext == ".svh" || ext == ".vh";
 }
 
+void Indexer::onWorkspaceDidChangeWatchedFiles(const lsp::DidChangeWatchedFilesParams& params) {
+    IndexGuard guard(*this);
+
+    std::vector<fs::path> pathsToAdd;
+
+    for (const auto& change : params.changes) {
+        fs::path path = change.uri.path();
```
evanwporter: Its a super simple fix. Just need to change this line too:

```
fs::path path = change.uri.getPath();
```

Honestly this might've been my bad. When I wrote `URI` I meant for `URI::path()` to refer to the underlying URI fomat:

https://github.com/hudson-trading/slang-server/blob/58408608018720b8fc83d828bb05eb73edc29297/include/lsp/URI.h#L9-L20

But it was never really meant to convey that this is a usable filesystem path. Future mistakes like this could be avoided by making it private, but I don't think that's the solution since access to each part of the URI is important.

evanwporter: windows tests should all pass now.

evanwporter: @AndrewNolte I was confused why in the other PR you it sounded like you didn't see these comments, but I didn't realize I had to submit the review :facepalm:

AndrewNolte: Oh makes sense, that always gets me too. I thought maybe I missed this bc it says '4 days ago' even though it was just submitted.

PR #258, opened 2026-02-20
Added basic URI unit tests

evanwporter:
Added a couple of basic URI tests to ensure basic functionality and verify that windows UNC style paths are parsed correctly.

### DIFF
--- a/tests/cpp/URITests.cpp
+++ b/tests/cpp/URITests.cpp
@@ -16,7 +16,7 @@ TEST_CASE("URI EmptyInput") {
     URI u("");
 
     // TODO: Right now it returns "/" as the URI string.
-    // It should return "/"
+    // It should return ""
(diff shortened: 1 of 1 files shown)


tests/cpp/URITests.cpp
```diff
@@ -0,0 +1,64 @@
...
+#else
+    CHECK(u.getPath() == R"(\x\ABC\ y.z)");
+#endif
+}
+
+TEST_CASE("URI EmptyInput") {
+    URI u("");
+
+    // TODO: Right now it returns "/" as the URI string.
+    // It should return "/"
```
AndrewNolte: I don't really get this comment?

evanwporter: Mb this is a typo. Ideally URI should return an empty output when it's passed an empty input, but currently it returns a "/".

PR #261, opened 2026-02-21
[VSCode] Add deprecation warning to `server.json` schema

evanwporter:
Been using indexGlob and I didn't even realize it was deprecated. Hopefully this will help future users from making the same mistake and allow it to eventually be phased out in a future release

Neovim may need to be changed as well

### DIFF
--- a/clients/vscode/resources/config.schema.json
+++ b/clients/vscode/resources/config.schema.json
@@ -11,10 +11,12 @@
         },
         "indexGlobs": {
           "type": "array",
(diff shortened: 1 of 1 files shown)


evanwporter:
Imo opinion I think most people who have repo with multiple cores are going to be using command files rather than indexing everytime they load up their project (like compile_commands.json), but the average person probable only has a single working dir (and maybe an include dir), so having this list with only one item is aesthetically unpleasing to the eyes lol. What about allowing both options? ie: specifying an array of items OR the items them selves

PR #262, opened 2026-03-01
Don't provide inlay hints when passed name == function arg name

evanwporter:
if a passed arg has the name as the function argument, then we don't provide a inlay hint

For example say we have a funtion:

```
function automatic logic do_something(logic a);
```

If we use this function usually the inlay hint would provide:

```
do_something(a: a);
```

Now because `a == a`. There's no hint:

```
do_something(a);
```


evanwporter:
alright this should be good.

src/document/InlayHintCollector.cpp
```diff
@@ -55,6 +55,15 @@ void InlayHintCollector::handle(const HierarchyInstantiationSyntax& syntax) {
                 if (paramIndex >= def.parameters.size()) {
                     break;
                 }
+
+                const auto name = def.parameters[paramIndex].name;
+                std::string passedArgName = paramSyntax->toString();
+                ltrim(passedArgName);
```
AndrewNolte, line 71: We can check that it's an id token, and compare the raw text. We'd also want to check that it's not behind a macro loc as well, since there's an edge case with something like this
```
`def FOO bar

my_func(`FOO);
```

src/document/InlayHintCollector.cpp
```diff
@@ -112,13 +121,24 @@ void InlayHintCollector::handle(const HierarchyInstantiationSyntax& syntax) {
                     if (portIndex >= ports.size()) {
                         break;
                     }
+
+                    const auto name = ports[portIndex]->name;
+                    std::string passedArgName = portSyntax->toString();
+                    ltrim(passedArgName);
```
AndrewNolte, line 139: Same here, also it's sort of repeat code that can be put in a lambda

src/document/InlayHintCollector.cpp
evanwporter: alright I'll try and make these changes before may XD

src/document/InlayHintCollector.cpp
```diff
@@ -55,13 +55,24 @@ void InlayHintCollector::handle(const HierarchyInstantiationSyntax& syntax) {
                 if (paramIndex >= def.parameters.size()) {
                     break;
                 }
+
+                const auto name = def.parameters[paramIndex].name;
```
AndrewNolte, line 69: We only need one paramIndex++, could be in this expr or right after.

AndrewNolte:
This needs a rebase/conflict fix now

PR #282, opened 2026-03-30
fix `ext.FindFiles` on windows

evanwporter:
`etx.find_files` is broken on windows. When you use the glob pattern to find the files, it says no files found. The issue I think is with this path:

```
      } else if (useGlob) {
        ret = (await glob(path.join(ws, str))).map((p: string) => vscode.Uri.file(p))
```

The fix is using the builtin vscode `findFiles` function 

@AndrewNolte is there a particular reason why we must have a separate find files func?

### DIFF
--- a/clients/vscode/src/extension.ts
+++ b/clients/vscode/src/extension.ts
@@ -23,7 +23,13 @@ import {
 import { PathConfigObject } from './lib/pathConfig'
 import { ProjectComponent } from './sidebar/ProjectComponent'
 import * as slang from './SlangInterface'
-import { AnyVerilogLanguages, anyVerilogSelector, getWorkspaceFolder, HDLFiles } from './utils'
+import {
+  AnyVerilogLanguages,
+  anyVerilogSelector,
+  getWorkspaceFolder,
+  HDLFiles,
+  toPosix,
+} from './utils'
 import { glob } from 'glob'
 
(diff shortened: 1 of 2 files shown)


clients/vscode/src/extension.ts
```diff
@@ -413,31 +413,6 @@ File input is sent to stdin, and formatted output is read from stdout.',
...
-      return []
-    }
-
-    // TODO: maybe use this.slangConfig.exclude.excludeDirs
-    // Or just wait for slang-format
-
-    const find = async (str: string): Promise<vscode.Uri[]> => {
-      let ret: vscode.Uri[]
-      if (path.isAbsolute(str)) {
-        ret = (await glob(str)).map((p: string) => vscode.Uri.file(p))
```
AndrewNolte, line 429: This is one reason- absolute paths don't work with vscode.workspace.findFiles. Another reason iirc is that git ignored files don't work with that method. Do you mind fixing this function instead? I think the useGlob branch is always taken now because of the gitignore issue.

evanwporter: alr i looked into it a little more and the issue was with glob expecting posix style path instead of the windows paths

evanwporter: should be good to go now

PR #287, opened 2026-04-01
Gray out Inactive Preprocessor Code

### LINKED ISSUE #267: [Feature] Gray out Inactive Preprocessor Code
Like C/C++ extension does for C/C++
Image

evanwporter:
Closes #267.

Slang stores the disabled tokens for all conditional directives. This PR takes advantages of that by collecting these token ranges into an array. This range will be exposed to vscode, which will then decorate them see:

* https://github.com/clangd/vscode-clangd/pull/193/changes
* https://clangd.llvm.org/extensions#inactive-regions
* https://reviews.llvm.org/D143974
* https://github.com/llvm/llvm-project/commit/3f6a904b2f3d8e974b223097956bb1ea51822782

On the client side, support for this feature is enabled via the LSP experimental client capabilities:

https://github.com/evanwporter/slang-server/blob/9da33e16a7c80e088fdbfe61c5a0a5a541bb95a0/include/lsp/LspTypes.h#L3504-L3505

The payload from the server to the client detailing the inactive regions for a file looks like:

```
{
  "uri": "file:///path/to/file.sv",
  "regions": [
    {
      "start": { "line": WW, "character": XX },
      "end":   { "line": YY, "character": ZZ }
    }
  ]
}
```

This PR only adds support to VSCode.

### DIFF
--- a/clients/vscode/src/extension.ts
+++ b/clients/vscode/src/extension.ts
@@ -274,9 +274,6 @@ File input is sent to stdin, and formatted output is read from stdout.',
     }
   }
 
-  /// Taken from vscode-clangd with slight modifcations see license at:
-  /// https://github.com/clangd/vscode-clangd/blob/master/LICENSE
-
   private async checkForUpdates(installedVersion: string | null): Promise<void> {
     const updated = await this.path.maybeInstallUpdate(this.context, this.logger, installedVersion)
     if (!updated) {
--- a/clients/vscode/src/SlangInterface.ts
+++ b/clients/vscode/src/SlangInterface.ts
@@ -4,6 +4,12 @@ import { Config } from './config.gen'
 
 export type { Config }
 
+export type ExperimentalCapabilities = {
+  inactiveRegions?: {
+    inactiveRegions: boolean
+  }
+}
+
 export enum SlangKind {
   Instance = 'Instance',
   Scope = 'Scope',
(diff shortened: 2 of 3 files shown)


AndrewNolte:
I would make a new LspExtensions file, since these technically aren't in the lsp. The stubs were generated from the lsp metamodel- https://github.com/AndrewNolte/lsprotocol/tree/reflect-cpp-stubs

evanwporter:
But i'll look into what clangd does.

evanwporter:
oh yeah also current method requires a token at the end for the disabled tokens to get attached too.

clients/vscode/src/extension.ts
```diff
@@ -268,6 +291,24 @@ File input is sent to stdin, and formatted output is read from stdout.',
...
+    if (!ranges) return
+    vscode.window.visibleTextEditors.forEach((e) => {
+      if (!this.inactiveDecorationType) return
+      if (e.document.uri.toString() !== fileUri) return
+      e.setDecorations(this.inactiveDecorationType, ranges)
+    })
+  }
+
+  private readonly inactiveDecorationType = vscode.window.createTextEditorDecorationType({
+    isWholeLine: true,
```
AndrewNolte: Could we try an inline ifdef/else? I think that case would look weird because of this arg.

evanwporter:
Yep it works

PR #288, opened 2026-04-01
feat: implement basic semantic tokens

### LINKED ISSUE #268: Implement Semantic Tokens
[Semantic Tokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens) provide syntax highlighting via the lsp; syntax highlighting is currently handled client side. This can also provide info for greyed-out code in untaken ifdef branches.

A basic first implementation can map slang token kinds to the lsp token kinds, which should definitely be more accurate than the current client-side implementations.

These are even more useful when they provide actual semantic info, for example highlighting parameters differently than variables, or instance variables differently from local variables.

AneesPatel:
**Summary**
* Implements basic Semantic Tokens support (fixes #268). This maps core Slang tokens to the standard LSP legend using a high-performance, single-pass delta encoder.

**Testing & Quality**
* Verified mapping logic with 10 core unit tests covering keywords, literals, macros, and operators.
* Formatted via `clang-format` to match project style guidelines.


src/document/SemanticTokens.cpp
```diff
@@ -0,0 +1,102 @@
...
+            if (!text.empty())
+                return SemanticTokenType::Operator;
+            return std::nullopt;
+        }
+    }
+}
+
+std::vector<uint32_t> computeSemanticTokenData(const SyntaxIndexer& indexer,
+                                               const SourceManager& sm) {
+    std::vector<uint32_t> data;
```
AndrewNolte, line 66: let's reserve() 5*collected

src/document/SemanticTokens.cpp
```diff
@@ -0,0 +1,102 @@
...
+        case TK::Directive:
+        case TK::MacroUsage:
+        case TK::MacroQuote:
+        case TK::MacroTripleQuote:
+        case TK::MacroEscapedQuote:
+        case TK::MacroPaste:
+            return SemanticTokenType::Macro;
+
+        default: {
+            auto text = parsing::LexerFacts::getTokenKindText(kind);
```
AndrewNolte: I would use `SyntaxFacts::getUnaryPrefixExpression` and `SyntaxFacts::getBinaryExpression` rather than making a string.

src/document/SemanticTokens.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+#include "slang/parsing/LexerFacts.h"
+#include "slang/parsing/Token.h"
+
+namespace server {
+
+using namespace slang;
+using TK = parsing::TokenKind;
+
+lsp::SemanticTokensLegend getSemanticTokensLegend() {
```
AndrewNolte: I would say to use `SemanticTokenTypes` in LspTypes.h, but that's generated as an `rfl::Literal`. Let's use slang's `SLANG_ENUM` to have the toString's generated and not have to maintain these two lists. We can also make this constexpr.

src/document/SemanticTokens.cpp
AndrewNolte: Actually this would be smaller, but let's have some heuristic proportional to collected. It eventually should be all of collected

src/document/SemanticTokens.cpp
```diff
@@ -0,0 +1,102 @@
...
+        return SemanticTokenType::Keyword;
+
+    switch (kind) {
+        case TK::Identifier:
+            return SemanticTokenType::Variable;
+
+        case TK::SystemIdentifier:
+        case TK::UnitSystemName:
+        case TK::RootSystemName:
+            return SemanticTokenType::Function;
```
AndrewNolte, line 32: This would require looking at the syntax to catch other funcitons, and looking at the AST for functions without the use of parens; Let's keep this PR syntax based, but this might present a regression compared to textmate.

PR #297, opened 2026-04-07
Support Non-ANSI style hovers

evanwporter:
~~Unfortunately~~ some people prefer the non-ANSI style of ports so this PR adds input/output declarations to the hovers to make it easier to see which ports to use.

See this reddit thread for the difference:

https://www.reddit.com/r/Verilog/comments/1jpz9w9/verilog_port_styles_ansi_vs_nonansi_which_one/

### DIFF
--- a/include/Hovers.h
+++ b/include/Hovers.h
@@ -9,7 +9,8 @@
 
 namespace server {
 
-lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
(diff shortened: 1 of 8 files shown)


AndrewNolte:
Yeah I would say so

evanwporter:
what?

evanwporter:
@AndrewNolte How much more work does your draft need? This PR is ~50% of the way there and I would find this feature helpful in my own work

PR #314, opened 2026-04-17
Improve LazyVim install instructions

matiasilva:
Hi there!

Many thanks for the excellent LSP server, I've found this a pleasure to use in my daily use. I run LazyVim and noticed that the installation instructions could be made a bit more idiomatic to match the existing Lazy plugin system.

The idea is to separate the nvim-lspconfig spec, which I presume will eventually make its way into the registry, from the plugin being enabled on LazyVim startup. When `slang-server` gets added to the Mason registry we can even get rid of the `mason = false`. While there are now two files, the line count is much lower and overall the setup is cleaner.

Thanks!

### DIFF
--- a/docs/start/installing.md
+++ b/docs/start/installing.md
@@ -82,7 +82,7 @@ return {
 }
 ```
 
-The above assumes that a slang-server config has been added to `nvim-lspconfig` using one of the preceding steps.
+The above assumes that a slang-server config has been added to `vim.lsp.config` or `nvim-lspconfig` using one of the preceding steps.
 
 Optionally, run `:LspInfo` to make sure the LSP was correctly installed.
 
--- a/docs/start/installing.md
+++ b/docs/start/installing.md
@@ -1,6 +1,5 @@
 # Installing
 
-
 ### Vscode
 
 Install the extension [here](https://marketplace.visualstudio.com/items?itemName=Hudson-River-Trading.vscode-slang). Then, it will prompt you to allow the extension to autoinstall the server binary from the [releases](https://github.com/hudson-trading/slang-server/releases) page. Alternatively, you can set `slang.path` to you built slang-server binary.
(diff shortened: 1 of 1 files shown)


docs/start/installing.md
```diff
@@ -52,49 +52,42 @@ if not configs.slang_server then
 end
 ```
 
-For users of lazy.nvim, the above could be added to their `nvim-lspconfig` spec at `~/.config/nvim/lua/plugins/nvim-lspconfig.lua` like this:
+For users of lazy.nvim, add the `nvim-lspconfig` config to `<runtimepath>/lsp/slang.lua`:
```
sgaulter: Let's keep the name as `slang_server`, as `slang` is already used by https://shader-slang.com/ LSP. This will hopefully avoid conflicts when `slang-server` is added to Mason repos.

docs/start/installing.md
```diff
@@ -52,49 +52,42 @@ if not configs.slang_server then
...
-              end,
-            },
-          }
-        end
-      end,
-    },
     servers = {
-      slang_server = {
-        enabled = true,
+      slang = {
```
sgaulter: As above.

sgaulter:
Hi Matias, thanks for this. Worth noting though that adding configs to the `<runtimepath>/lsp` directory requires Neovim ≥ v0.11 as well I believe. The previous incantation is generic enough to be used by Neovim < v0.11 as well.

We're in danger of ending up with an MxN problem of documenting for Neovim versions and package managers.

I agree using the `lsp` directory is cleaner when it's available. Perhaps it should be moved to just after the `vim.lsp.config` method, making it clear it's an alternative when using ≥ v0.11.

Then keep the new, condensed `LazyVim` enable instructions, assuming the config has been added with one of the preceding methods.

matiasilva:
Thanks for the review Simon, definitely agree with all. How's it looking now?

docs/start/installing.md
```diff
@@ -52,49 +66,30 @@ if not configs.slang_server then
...
-        enabled = true,
+        -- Tell LazyVim that Mason isn't needed since this is a manual config
         mason = false,
       },
     },
   },
 }
 ```
 
+The above assumes that a slang-server config has been added to `nvim-lspconfig` using one of the preceding steps.
```
sgaulter:
```suggestion
The above assumes that a slang-server config has been added to `vim.lsp.config` or `nvim-lspconfig` using one of the preceding steps.
```

sgaulter:
Great, thanks for the changes. Just one minor comment, but happy to merge otherwise.

PR #315, opened 2026-04-19
feat: show elaborated parameter value on hover

wasinsangdam:
Description:                                                                 

## Summary                                                                   

Adds an elaborated `Value:` line to parameter hovers so users can see what a parameter actually resolves to in the loaded design, not just its declared default. This is especially useful for parameters that are passed down through several module layers via overrides.
                                                                     
The value is resolved server-side in two ways, in order of preference:       

1. **Active instance** — the VSCode Hierarchy View now reports the currently focused instance to the server via a new `slang.setActiveInstance` request. When the hovered parameter belongs to that instance's module, its instance-specific value is shown.                                         
2. **Module-based fallback** — when no active instance is set, or it does not match the hovered module, the server walks every instance of the module via `moduleToInstances`. If they all share the same value the hover shows it;  otherwise the value is treated as ambiguous and omitted, leaving the existing default-value hover unchanged.                                   

This means the feature works out of the box even without any client-side state: as soon as the design is elaborated, hovers on parameters of single-instance modules (or modules whose instances happen to agree) light up.                                                                          

- `ServerCompilation::getElaboratedParamValue(moduleName, paramName, activeInstancePath?)` encapsulates the lookup logic.
- `ServerDriver::getDocHover` calls it for `ParameterSymbol` hovers and appends                                                              a `Value:` line when a value is returned.                                  
- New LSP request `slang.setActiveInstance` (string `hierPath`) lets the client push the focused hierarchical instance to the server. Stored on `SlangServer::m_activeInstancePath`.                                       
- VSCode extension: `ProjectComponent` calls `setActiveInstance` whenever the Hierarchy View focus changes.                                              

## Tests                                                                     

- `HoverParameterElaboratedUniqueInstance` — single-instance module, elaborated                                                                 value surfaces without any active-instance state.
- `HoverParameterElaboratedAluModule` — multi-instance ALU under CPU; edits  `cpu_testbench.sv` to set `DATA_WIDTH=40`, saves to trigger `comp->refresh()`, and verifies that `alu.sv`'s `WIDTH` hover shows `40` (not the declared default of `32).

All 947 assertions / 185 cases in `server_unittests` pass.                   

## Notes for reviewers

- The active-instance path is purely additive — clients that don't know about `slang.setActiveInstance` just get the module-based fallback behavior.     
- The elaborated value depends on the full design compilation, which only refreshes on `didSave`. This is worth keeping in mind when manually testing.

### DIFF
--- a/include/SlangServer.h
+++ b/include/SlangServer.h
@@ -68,6 +68,9 @@ protected:
     // The waveform viewer client
     std::optional<waves::WcpClient> m_wcpClient = std::nullopt;
 
+    // The hierarchical path of the currently active instance, set via slang.setActiveInstance
+    std::optional<std::string> m_activeInstancePath;
+
 public:
     SlangServer(SlangLspClient& client);
 
@@ -134,6 +137,9 @@ public:
 
     // Add a -D define to .slang/local/server.json and reload config
     std::monostate addDefine(const std::string& macroName);
+
+    // Store the active instance path from the Hierarchy View for enriched hover
+    std::monostate setActiveInstance(const std::string& hierPath);
 
     ////////////////////////////////////////////////
     /// Server Lifecycle
--- a/clients/vscode/src/SlangInterface.ts
+++ b/clients/vscode/src/SlangInterface.ts
@@ -144,6 +144,11 @@ interface ExpandMacroArgs {
 export async function expandMacros(args: ExpandMacroArgs): Promise<boolean> {
   return await vscode.commands.executeCommand('slang.expandMacros', args)
 }
+
+/// Notify the server of the currently active hierarchical instance for enriched hover
+export async function setActiveInstance(hierPath: string): Promise<void> {
+  await vscode.commands.executeCommand('slang.setActiveInstance', hierPath)
+}
 ////////////////////////////////////////////////////////////
 /// server -> client is in commands in the project component
 ////////////////////////////////////////////////////////////
--- a/clients/vscode/src/sidebar/ProjectComponent.ts
+++ b/clients/vscode/src/sidebar/ProjectComponent.ts
@@ -639,6 +639,7 @@ export class ProjectComponent
       }
 
       this.focused = instance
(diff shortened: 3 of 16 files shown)


AndrewNolte:
I think it'd be good to have an active instance per module, rather than a single active instance. And there should be some indicators of the active instance-
- The modules view should indicate which instance is active. I also think one should be selected by default- probably the first one / highest in the hierarchy.
- An inlay hint after the module name, which is the selected hierarchical path.

include/SlangServer.h
```diff
@@ -68,6 +68,9 @@ class SlangServer : public lsp::LspServer<SlangServer>, public SlangServerWcp {
     // The waveform viewer client
     std::optional<waves::WcpClient> m_wcpClient = std::nullopt;
 
+    // The hierarchical path of the currently active instance, set via slang.setActiveInstance
+    std::optional<std::string> m_activeInstancePath;
```
AndrewNolte, line 72: This should be a map and live in the server compilation

wasinsangdam: Thanks for the feedback! Quick clarification on the inlay hint: should it be a standard LSP inlay hint rendered in the editor (e.g., virtual text after `module alu` in alu.sv showing the active instance path), or text added to the module label in the sidebar Modules view? Happy to implement either — just want to make sure I build the right one.

Will go ahead and implement the per-module active instance map and the default selection + indicator in the Modules view in the meantime.

AndrewNolte: Actually I have a big refactor of the client/server api, since right now the source locations can get out of sync with any edits. It also touches a lot of the server compilation code, so let me just add this feature onto that pr. Thanks for giving this a shot though! Once that's more stable I think there will be a lot of interesting features to contribute there, like displaying signal values in the inlay hints over wcp.

When I was experimenting with it I found codelenses a lot better actually, since then it won't clip the instance path. I also added a status bar item to indicate the selected instance.

AndrewNolte:
Implemented as part of https://github.com/hudson-trading/slang-server/pull/538

Issue #310, opened 2026-04-16
Incorrect Hover Markdown for Backticks


henry-hsieh:
Slang-Server Version: 0.2.3+34e8314
Slang Version: 10.0.134+1926b2246

Failed code snippets:

```systemverilog
`define MACRO_A 10
`define JOIN_MACRO(name) name```MACRO_A
```
The hover of `MACRO_A` is:
```lua
{
  jsonrpc = "2.0",
  result = {
    contents = {
      kind = "markdown",
      value = "**Token:** `Identifier`  \n **DefineDirective** : ``define MACRO_A 10`  \n **CompilationUnit** : `\n`  \n"
    }
  }
}
```
The hover of `JOIN_MACRO` is:
```lua
{ 
  jsonrpc = "2.0",
  result = {
    contents = {
      kind = "markdown",
      value = "**Token:** `Identifier`  \n **DefineDirective** : `\n`define JOIN_MACRO(name) name```MACRO_A`  \n **CompilationUnit** : `\n`  \n"
    }
  }
}
```

Because valid SystemVerilog syntax will have at most 3 consecutive backticks, we should wrap the inline code span with 4 backticks with one space to quote the value:
`````markdown
```` `define MACRO_A 10 ````
```` `define JOIN_MACRO(name) name```MACRO_A ````
`````
They should be correctly rendered to these:
```` `define MACRO_A 10 ````
```` `define JOIN_MACRO(name) name```MACRO_A ````

As for the language code block, it only needs to use 3 backticks because markdown only recognizes the consecutive backticks at line start as the fence of code block:
````markdown
```systemverilog
// ``` backticks in middle of line
logic [0:0] clk;
```
````
This should be rendered as:
```systemverilog
// ``` backticks in middle of line
logic [0:0] clk;
```

However, because slang-server actually extract the comment as the markdown content, we should be very careful about the previous pattern may mess up the markdown content:

`````markdown
**Variable** `clk` in `test`
Type: `logic[0:0]` 


---

``` backtick in the middle

---

````systemverilog
logic [0:0] clk;
````
`````
Here is how it looks in GitHub markdown:

<img width="1853" height="504" alt="Image" src="https://github.com/user-attachments/assets/21198227-fa28-4462-b237-32f6880b901e" />

---

In summary, I have 3 suggestions:

1. Use 4 backticks with 1 space to quote the inline rendered value for fixing the rendering issue of macro.
[...]

PR #317, opened 2026-04-20
Add plain text hover option (escape markdown related symbols)

evanwporter:
This PR adds the ability for the user to change how documentation comments are displayed. One option is to display `markdown`. The other option is to display `plaintext` (ie: text except all the markdown related symbols are escaped).

The markdown symbol recognition is taken from clangd:

https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/support/Markup.cpp

This feature mirrors the documentation config in clangd;

https://clangd.llvm.org/config#documentation

Part 1 of a fix for #310.

Couple of things to think about:

(1) Clangd keeps spacing on the left while slang-server trims it. `escapeMarkdownLine` would need to be modified if we ever decided to respect this spacing. Example of what I mean:

`//              COMMENT` <- clangd wouldn't trim all that extra spacing.

(2) The LLVM adapted functions are left in the style of the LLVM repo (ie: var names start w/ uppercase).

(3) The JSON golden test generator doubles the number of backslash so its JSON compliant.

(4) Clangd has `plaintext` by default whereas this pr is `markdown` by default.

(5) Clangd escapes html tags regardless of the setting. This PR escapes html tags only when `plaintext` is set.


evanwporter:
Literally none of the code so far is my own.

evanwporter:
https://github.com/llvm/llvm-project/blob/main/clang-tools-extra/clangd/unittests/SymbolDocumentationTests.cpp

evanwporter:
In the future a full on doxygen parser would be cool for the hovers. Or at least displaying simple tags like `@brief` in an aesthetically pleasing way.

PR #323, opened 2026-04-29
[VSCode] don't reveal hierarchy on goto signal waveform

evanwporter:
If I'm in the vaporview tab, then I want to stay in the vaporview tab.

### DIFF
--- a/clients/vscode/src/sidebar/ProjectComponent.ts
+++ b/clients/vscode/src/sidebar/ProjectComponent.ts
@@ -1101,7 +1101,7 @@ export class ProjectComponent
         await this.openBuildFile({ name: basename, top: top })
       }
       await this.setInstance.func(fullpath, {
-        revealHierarchy: true,
+        revealHierarchy: false,
         revealFile: true,
         revealInstance: true,
         focus: 'editor',


AndrewNolte:
This is specifically from the instances/modules view, there's another for the hierarchy view. 

But I also don't understand the comment "If I'm in the vaporview tab...". In this flow, you are starting out in the slang instances tab. And it opens the waveform after this call. I think you meant to change `showInEditorFromNetlist` and `showInEditorFromVaporview`? Or maybe you're asking to also show the netlist tab? In which case you may need another call to the vaporview api.

evanwporter:
I’ll send a GIF in a bit.

evanwporter:
@AndrewNolte 

https://github.com/user-attachments/assets/83d8790d-855d-4896-8e4b-972ee98a53a0

I don't like how the hierarchy pops up. 

Also another issue is that verilator often groups everything under a TOP specifier, so when using verilator generated vcd waveforms it breaks unfortunately.

evanwporter:
Yep, my mistake. When I was searching for the problem, I made a few different changes and then I undid the non-fixes and committed my fix. But unfortunately, it seems I committed the wrong change.

AndrewNolte:
I believe the TOP thing happens only when your top level has i/o

PR #328, opened 2026-05-04
Don't ignore neovim client for CI

toddstrader:
`build.yml` is useful for neovim client changes since we have CI for it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ on:
     branches: [main]
     paths-ignore:
       - 'docs/**'
-      - 'clients/**'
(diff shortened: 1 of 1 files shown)


toddstrader:
Land #330 first to fix the CI.

sgaulter (approved):

.github/workflows/build.yml
```diff
@@ -18,15 +18,15 @@ on:
     branches: [main]
     paths-ignore:
       - 'docs/**'
-      - 'clients/**'
+      - 'clients/vscode/**'
```
AndrewNolte, line 21: Might be easier to switch to 'paths:' instead of ignores

PR #330, opened 2026-05-04
Fix missing diags

toddstrader:
Fixes `test_diags_update` which was failing because dirty URIs were not being republished on `textDocument/publishDiagnostics`.


toddstrader:
I'm confused now.  I thought this test was failing on `main`, but I see now that the red is due to the coverage job.  I do, however see the failure on recent PRs I put up.  E.g.:
https://github.com/hudson-trading/slang-server/actions/runs/25328695764/job/74256288672

Also, I don't know why this only impacts the clang build.

PR #331, opened 2026-05-05
Additional neovim compatibility tests

toddstrader:
In addition to checking that the server handles commands immediately before dispatching them, plugin commands test for the entire set of necessary commands so as to prevent weird behavior in the case of incompatibility.  Also checks the semantic version (will need to be manually bumped).

### DIFF
--- a/clients/neovim/lua/slang-server/_commands/setTopLevel.lua
+++ b/clients/neovim/lua/slang-server/_commands/setTopLevel.lua
@@ -5,10 +5,15 @@ local M = {}
 ---@type slang-server.ui.Subcommand
 M.setTopLevel = {
    impl = function(args, opts)
+      local capabilities = require("slang-server._lsp.capabilities")
+      local bufnr = vim.api.nvim_get_current_buf()
+      if not capabilities.check_or_notify(bufnr, { "slang.setTopLevel" }) then
+         return
+      end
+
       local client = require("slang-server._lsp.client")
       local handlers = require("slang-server.handlers")
 
-      local bufnr = vim.api.nvim_get_current_buf()
       local file = args[1] or vim.api.nvim_buf_get_name(bufnr)
       client.setTopLevel(bufnr, handlers.defaultHandlers, { uri = file })
    end,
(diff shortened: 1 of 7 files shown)


toddstrader:
Land #332 before this one to fix the CI.

clients/neovim/lua/slang-server/_commands/setTopLevel.lua
```diff
@@ -5,10 +5,15 @@ local M = {}
 ---@type slang-server.ui.Subcommand
 M.setTopLevel = {
    impl = function(args, opts)
+      local capabilities = require("slang-server._lsp.capabilities")
+      local bufnr = vim.api.nvim_get_current_buf()
+      if not capabilities.check_or_notify(bufnr, { "slang.setTopLevel" }) then
+         return
+      end
+
```
sgaulter, line 13: I feel like these explicit checks (and maybe even `check_or_notify`) shouldn't be necessary. We could just let `client`'s `lsp_execute` perform the check before issuing each command and fire its failure handler with the error message if it fails.

clients/neovim/lua/slang-server/_lsp/client.lua
```diff
@@ -9,20 +11,9 @@ local lsp_execute = function(bufnr, params, handlers)
...
-            end
-         end
-      end
-   end
-
-   if not client_found then
-      on_failure("No client found")
+   local ok, err = capabilities.command_supported(bufnr, command)
+   if not ok then
+      on_failure(err)
```
sgaulter, line 16: I don't suppose `command_supported` adds a huge amount of processing on every LSP call, but as we'd expect it be totally static it would be a prime candidate to memoise.

clients/neovim/lua/slang-server/_commands/setTopLevel.lua
toddstrader: I can backpedal on this, but my thinking was that it would be better to tell the user immediately at the point of executing the plugin command that something isn't supported as opposed to getting part way into the plugin command, discovering an LSP command isn't there and leaving the user in some wonky state.

Also, the downside to checking like this is that we'll need to remember to update the list.  So there's that.

clients/neovim/lua/slang-server/_lsp/client.lua
toddstrader: Done.

PR #334, opened 2026-05-06
Add new builds/releases for mason; Clean up build matrix

toddstrader:
- Names build entries so it's easier to read in github (release binary names should be unchanged)
- Adds arm builds for all platforms (other than mac)
- Decided against adding intel mac since it's on the way out


AndrewNolte:
I don't think mason actually requires you to support every combo in the matrix, so I don't think it's worth having intel mac. Arm mac runners are already so unavailable already- intel macs would be even worse.

PR #335, opened 2026-05-06
neovim: uri_to_fname()

toddstrader:
Direct URI to File path conversion does not work on windows.


toddstrader:
Note to self: does this fix neovim + windows CI?

toddstrader:
OK, yeah.  I think this would be a problem but we've got more fundamental problems with the neovim GH action.  However, said action does appear to run fine on windows when the action is tested by itself:
https://github.com/luarocks/luarocks/actions/runs/24047544273/job/70134660683
Some CI is red, but I think the parts of the matrix we care about work.  So I can look at that separately.  But I don't think it's related to this PR.

PR #336, opened 2026-05-08
Fix macro completion label formatting

evanwporter:
Consider I have this:

```
`<cursor here>
```

Then list of macros pops up. This is great.

I select my macro.

```
``MY_MACRO<cursor here>
```

Uh oh. Now I need to go back and remove a backtick--well not anymore with this pr


evanwporter:
ayo nvm

PR #337, opened 2026-05-10
Vscode grammar: Highlight DPI import/export declarations

### LINKED ISSUE #322: dpi highlights
Image

is the bright red highlighting intended? Imo its a little out of place when compared to every other highlight--which tends to be darker; more modest coloring.

I think the regular bold black would look better

nahomar:
Closes #322.

## What

Adds a `dpi-import-export` pattern to the SystemVerilog tmLanguage so `import "DPI-C"` / `export "DPI-C"` (and the legacy `"DPI"`) declarations get the keyword/string/modifier highlights instead of falling through to the default scope.

Covers the full IEEE 1800 §35 grammar:

- `import "DPI-C" [pure|context] [c_id =] function|task ...`
- `export "DPI-C" [= c_id] function|task name;`

| Token                | Scope                                  |
| -------------------- | -------------------------------------- |
| `import` / `export`  | `keyword.control.systemverilog`        |
| `"DPI-C"` / `"DPI"`  | `string.quoted.double.systemverilog`   |
| `pure` / `context`   | `storage.modifier.systemverilog`       |
| c_identifier alias   | `entity.name.function.systemverilog`   |
| `=` (in alias forms) | `keyword.operator.assignment.systemverilog` |

The trailing `function` / `task` keyword and the rest of the prototype are intentionally left to the existing `#function-task` rule, so a full declaration like `import "DPI-C" function int foo(input int x);` ends up fully tokenized end-to-end.

## Test plan

- New unit test `tests/grammar/unit/dpi.sv` with targeted scope assertions for plain / `pure` / `context` / legacy `"DPI"` / import-alias / `pure` + alias / export function / export task / export-alias.
- New snapshot test `tests/grammar/snap/dpi.sv` capturing the full token stream of all variants for visual review.
- `pnpm run test:grammar` — passes (3/3, including the new file).
- `pnpm run test:grammar:snap` — passes (5/5, including the new snapshot, no regressions in existing 4).


AndrewNolte:
Tested this out, it fixes the import/export coloring, but not the 'function' keyword.

AndrewNolte:
Please thoroughly test and review AI generated prs- it wastes my time and makes us less likely to look at PRs from authors that submit them.

AndrewNolte:
The function/task is still highlighted red because they're part of the language configuration, I'll fix that in a followup PR.

PR #339, opened 2026-05-10
Add hovers and completions for built-in system procedures

### LINKED ISSUE #270: [Feature Request] System task hover information
Currently, when hovering over a system task like $countones or $stable, nothing is shown. The idea would be that system tasks defined in 20.1 and 21.1 of the IEEE 1800 standard had a small description when you hovered over it. Although many of the names are self-explanatory, I think it would be a cool feature (maybe I just like hover windows too much).

nahomar:
Closes #270.

> [!NOTE]
> Built on top of [#338](https://github.com/hudson-trading/slang-server/pull/338) (the Apple-clang 16 build fix). Once #338 lands I'll rebase to drop that commit; reviewers can ignore it here.

## What

System tasks like `\$display`, `\$bits`, `\$clog2` are resolved by slang via the `Compilation::getSystemSubroutine` registry, not the AST Symbol scope hierarchy. As a result, `getDefinitionInfoAt` was silently skipping them — hover, goto-def, and references all returned nothing for `\$xxx` tokens. This change wires them up and supplies docs for **every slang-known `KnownSystemName`**.

### Pieces

- **`SystemTaskDocs.{h,cpp}`** — a `constexpr std::array<SystemTaskDoc, 287>` indexed by slang's `KnownSystemName` enum. Each entry holds `signature`, `description`, and `ieeeSection` (the section reference is omitted for the few entries with no IEEE 1800 home).
- **`DefinitionInfo`** — gets a new `const SystemTaskDoc* sysTaskDoc` field plus a `bool sysIsTask` so the hover header can correctly render "System task" vs "System function" based on `SubroutineKind`.
- **`getDefinitionInfoAt`** (`src/ServerDriver.cpp`) — after the Symbol-lookup path fails, if the token starts with `\$` and resolves via `Compilation::getSystemSubroutine`, return a synthetic `DefinitionInfo` carrying the doc pointer.
- **`Hovers.cpp`** — short-circuits when `info.sysTaskDoc` is set and renders `**System task** \$name (IEEE 1800 §X)` followed by a code-block signature and the description paragraph.
- **`static_assert(countDocumented() == kNumKsn - 1)`** in `SystemTaskDocs.cpp` enforces that every `KnownSystemName` (other than the `Unknown` sentinel) has docs. Adding a new entry to slang's `systemnames.txt` upstream will fail the build until the docs row is added here — exactly per the maintainer's note in the issue thread.

## Coverage

All 286 named entries documented across:

- §16.9 / §16.16 SVA sampled-value & global-clock functions
- §16.13 Sequence methods (`triggered`, `matched`)
- §16.16 `\$global_clock`, `\$inferred_clock`, `\$inferred_disable`
- §6.16 / §6.19 / §7.9 / §7.10 / §7.12 Built-in array, queue, associative-array, enum, and string methods
- §18.7–18.9 Object randomization (`randomize`, `rand_mode`, `constraint_mode`)
- §18.13 Random (`\$random`, `\$urandom`, `\$urandom_range`)
- §20.2–20.16 Utility system tasks/functions (sim control, time, conversion, data/array query, math + hyperbolic + trig, bit-vector, severity, plus-args, assertion control, sampled value, coverage, distributions, queues)
- §21.2–21.7 I/O (display family with all base-suffix variants, file I/O including extras like `\$ferror`/`\$fread`/`\$fseek`, memory load incl. string-source `\$sreadmemb`/`\$sreadmemh`, PLA modeling, `\$timeformat`, VCD `\$dumpvars`/`\$dumpports` family)
- §22 Compiler/simulator interactive control

The `ieeeSection` field is empty for a handful of entries with no IEEE 1800 reference (`\$psprintf`, `\$stacktrace`, `\$getpattern`, `\$system`, `\$deposit`).

## Test plan

- New unit tests in [tests/cpp/HoverTests.cpp](tests/cpp/HoverTests.cpp):
  - `HoverSystemTask` — content/IEEE-section/signature for `\$display`/`\$finish`/`\$bits`/`\$clog2`.
  - `HoverSystemTask_FunctionVsTaskLabel` — verifies the header reads "System function" for `\$bits` and "System task" for `\$finish`.
  - `HoverSystemTask_NoFalsePositiveOnNonRegisteredDollarTokens` — `q[\$]` and `\$root` must not produce system-task hover.
- Updated golden files: `FindSymbolRef.out`, `FindSymbolRefMacro.out`, `LoadTransitivePackages.out`. Every diff is a new `^^^^^^ Sym \$X : Y` line — system-task tokens that the scanner previously couldn't resolve.
- `cmake --build build/macos-release -j8 --target server_unittests && build/macos-release/bin/server_unittests` — **191 test cases / 58254 assertions, all pass** (was 188/58235 before this change; the +3 are the new HoverSystemTask cases).
- `clang-format v20.1.8` clean on all changed C++ files.


src/ServerDriver.cpp
```diff
@@ -556,6 +558,25 @@ std::optional<DefinitionInfo> ServerDriver::getDefinitionInfoAt(const URI& uri,
     else {
         symbol = analysis->getSymbolAtToken(declTok);
         if (!symbol) {
+            // System tasks ($display, $finish, etc.) aren't materialized as
```
AndrewNolte: This branch could be syntax based (SystemNameSyntax), rather than in this else branch.

tests/cpp/HoverTests.cpp
```diff
@@ -157,3 +157,132 @@ endmodule
...
+        CHECK(content.value.find("Halts") != std::string::npos);
+    }
+
+    {
+        auto cursor = doc.before("$bits(int)");
+        auto hover = doc.getHoverAt(cursor.m_offset);
+        REQUIRE(hover.has_value());
+        auto content = rfl::get<lsp::MarkupContent>(hover->contents);
+        CHECK(content.value.find("$bits") != std::string::npos);
+        CHECK(content.value.find("§20.6.2") != std::string::npos);
```
AndrewNolte: For opiniated outputs like these it's better to do a golden test. Right now you're assigning nameToken to the function call, but this is really more like a reference (It's `sym` rn in the golden tests). So let's have the name token be in an invalid buffer. Ideally DefinitionInfo will be changed to a tagged variant in the future.

AndrewNolte:
Not sure why github failed to mark it as co-authored, and also managed to land the entire description when I had not intended for that.

PR #345, opened 2026-05-19
[VSCode Snippets] Use named end blocks instead of comments

evanwporter:
~~-Title generated by copilot~~

lgtm and release a new vscode ext version

neovim side might need to change too

### DIFF
--- a/clients/vscode/languages/sv/systemverilog.snippets.json
+++ b/clients/vscode/languages/sv/systemverilog.snippets.json
@@ -42,7 +42,7 @@
       "class ${1:className};",
       "\tfunction new();",
       "\t\t$0",
-      "\tendfunction // new()",
+      "\tendfunction : new",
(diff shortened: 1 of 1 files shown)


clients/vscode/languages/sv/systemverilog.snippets.json
```diff
@@ -42,8 +42,8 @@
       "class ${1:className};",
       "\tfunction new();",
       "\t\t$0",
-      "\tendfunction //new()",
-      "endclass //$1"
+      "\tendfunction // new()",
```
AndrewNolte: These can also be named end blocks

evanwporter: yeah, I thought about replacing it with a named end block, but I didn't because in this case the parenthesis might look better. Idk though

PR #347, opened 2026-05-19
Add `addModuleMemberKwCompletions` back

evanwporter:
At some point the primary function from #152 was removed.

There should've been a test to catch this.

This also fixes `always_comb`, `always_ff`, etc.


evanwporter:
I wonder if there's a reverse git blame--to see what got removed :hmm:

evanwporter:
need to update tests

AndrewNolte:
Addressing this in this PR - https://github.com/hudson-trading/slang-server/pull/363

Issue #346, opened 2026-05-19
Context Aware Keyword Completions


evanwporter:
https://github.com/hudson-trading/slang-server/blob/main/clients/vscode/languages/sv/systemverilog.snippets.json

Add function snippets. Shouldn't take too long--I just wasn't sure (a) why they weren't already added and (b) what the trigger words should be

AndrewNolte:
So the end state with snippets is having them in completions. It would be nice to have them mostly generated from syntax.txt, if you're open to working on that. Right now our neovim friends don't have much for snippets.

evanwporter:
> So the end state with snippets is having them in completions. It would be nice to have them mostly generated from syntax.txt, 

Oh, this sounds really cool. Would this be like #152?

> if you're open to working on that.

Maybe...I don't want to commit. Let me look at the lsp for completions.

> Right now our neovim friends don't have much for snippets.

If we do internalize the snippets then client-side snippets wouldn't be necessary so that'd be nice

Also, this is a side question, but what causes a `declType` error? See the formatting file:

```
template<bool ForHover>
std::string getTypeStringImpl(const ast::Type& declType) {
    if (declType.isError()) {
        return "Incomplete type";
    }
```

AndrewNolte:
Yeah, like #152. 

That case is for incomplete / error types. Like let's say some struct or typedef depends on a parameter that isn't set, or it set to an undefined define. It'll still create the type, but it's type would be set to error. Ideally we would try to show as much information as possible though, so like "Incomplete type because <x> is not complete here". We would want to wrap it in a "Bad Type", so we for example can at least see the struct members for completions. Slang does this wrapping with some forms of bad expressions at the moment, but not yet with types.

ting-hong-shieh:
Hi, I'd like to pick this up.

I checked the related work: #362 already provides VS Code-side `function` and `task` snippets, while the broader server-side completion work in #348 was closed before it was finished. Rather than revive the full scope of #348 at once, I propose a focused first PR that:

- adds context-aware server-side `function` and `task` snippet completions for `ModuleMember`;
- uses `MarkupContent` for the completion documentation;
- adds completion/golden-test coverage, including cases where these snippets should and should not appear; and
- keeps the existing client-side snippets for now, unless you would prefer them removed once the server-side completions are available.

This should bring the snippets to non-VS Code LSP clients while keeping the change reviewable. Does that scope sound right?

evanwporter:
[...]

PR #348, opened 2026-05-20
server side completions

evanwporter:
Following up with #346. This PR adds completions to the server side. Atm its currently a port of the vscode snippets.

Stacks onto #347 

Couple things to note:

(1) I removed the associative array completion because I felt like `AA` should be a snippet the user adds not slang

(2) I thought about parsing `syntax.txt` but I dismissed the idea because newlines, spaces, tabs, context and placeholders would all still have to be manually defined, so it would be much less work just to create a curated selection of completions.

   (a) If we do want to go this route then I think it'd be funny to give this work to a google summer of code student (idk how the program works).

(3) I created the yaml as a base to generate the C++ code. Eventually we want to move this into cmake and generate at configure time

(4) `gen_snips.py` is almost entirely written by AI-- I haven't looked over it, I just verified the output.

(5) The yaml is based off the textmate yaml

(6) I mentioned `context` in (2), what I mean by this is that each module needs to hold the context with which it's in. Ex: If I'm in a module already, then I don't need autocompletion results for module. Likewise, you don't need to create a package within a package. This still needs to be done.

(7) Idk how we will decide whether to add in parameter port list and imported module into a module declaration autocompletion--maybe put these at the end of the list (ie: $2, $3), and then the user can skip them? If these things are empty, then the formatter can strip it?


evanwporter:
I asked chatgpt to come up with a list of every completion that needs to be done based on syntax.txt. Here's what it came up with:

# Top-level / module-member declarations
- module
- interface
- program
- package
- import
- export
- parameter
- localparam
- type parameter
- genvar
- nettype
- typedef
- typedef enum
- typedef struct
- typedef struct packed
- typedef union
- typedef union packed
- struct
- struct packed
- union
- union packed
- enum
- logic
- bit
- reg
- wire
- tri
- integer
- int
- shortint
- longint
- byte
- time
- real
- shortreal
- realtime
- string
- chandle
- event

# Module / interface members
- assign
- defparam
- bind
- modport
- clocking
- default clocking
- default disable iff
- timeunit
- timeprecision
- specify
- primitive
- extern module
- extern interface method

# Instantiation / ports
- named port connection
- wildcard port connection
- ordered port connection
- parameter override
- named parameter override
- instance

# Generate constructs
- generate
- for generate
- if generate
- case generate
- begin generate block
- genvar declaration

# Procedural blocks
- initial
- final
- always
- always_comb
- always_ff
- always_latch

# Sequential / procedural statements
- begin end
- fork join
- if else
- case
- case inside
- forever
- repeat
- while
- do while
- for
- foreach
- return
- break
- continue
- wait
- wait fork
- wait_order
- disable
- disable fork
- randcase
- randsequence
- void cast call

# Functions / tasks
- function
- automatic function
- task
- automatic task
- extern function
- extern task
- pure virtual function
- class constructor

# Class declarations / members
- class
- virtual class
- interface class
- class extends
- class implements
- class extends implements
- function new
- constraint
- extern constraint
- pure constraint
- rand property
- static property
- local property
- protected property

# Assertions
- assert
- assume
- cover
- assert property
- assume property
- cover property
- cover sequence
- restrict property
- expect property
- disable iff
- property
- sequence
- checker
- let

# Constraints
- constraint block
- soft constraint
- unique constraint
- implication constraint
- if constraint
- foreach constraint
- disable soft
- solve before
- dist constraint

# Coverage
- covergroup
- covergroup with function sample
- coverpoint
- coverpoint iff
- cross
- bins
- illegal_bins
- ignore_bins
- wildcard bins
- binsof
- intersect

# DPI
- import DPI-C function
- import DPI-C task
- export DPI-C function
- export DPI-C task

# System tasks / common utility snippets
- display
- finish
- fatal
- error
- warning
- info
- queue
- mailbox
- associative array
- dynamic array

# Compiler directives
- define
- undef
- ifdef
- ifndef
- elsif
- else
- endif
- include
- timescale
- default_nettype
- resetall
- begin_keywords
- end_keywords
- celldefine
- endcelldefine
- pragma

for some of these entries its my first time seeing them (`pure constraint`?!?)

evanwporter:
<s>
@AndrewNolte Consider the following:

image

This is what it looks like when using the frontend vscode snippings.

Now see what it looks like when using the backend autocompletions:

image

The frontend client side snippings are faster on startup since they don't wait for the server to get spun up and they look significantly better. Assuming the aesthetics can't be replicated in the backend autocompletions, I think we should try and use the frontend snips over the backend snippets.

Oh yeah, the main issues aesthetically are that the hover doesn't display markdown, and it doesn't have the little label to the right.

Regarding neovim, we can allow backend snippings to be disabled via a client <--> server setting.
</s>

src/completions/SyntacticalCompletions.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include <string_view>
+
+namespace server::completions {
+
+void addModuleMemberKwCompletions(std::vector<lsp::CompletionItem>& results) {
+    for (const auto& snippet : SV_MODULE_MEMBER_SNIPPETS) {
+        results.emplace_back(lsp::CompletionItem{
+            .label = std::string(snippet.label),
+            .kind = lsp::CompletionItemKind::Snippet,
+            .documentation = snippet.documentation.empty()
```
AndrewNolte: They don't look nice because you're using a string instead of markup content- 
```
/**
	 * A human-readable string that represents a doc-comment.
	 */
	documentation?: string | [MarkupContent](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#markupContentInnerDefinition);
```

evanwporter: I see. This is very interesting...I made a bad assumption :(

evanwporter:
image

looks better now

evanwporter:
Hey Andrew,

Can you help me with this PR? I’m trying to speedrun it since I’m starting
a new job soon and want to finish it before. It’d be great if you could
build out the context class that you made earlier.

This is broken when the cursor is in the top level of a file but below a
module/interface/classes declaration; it should identify it as being in the
top level (via the compilation unit) but it actually says it’s somewhere
else.

You’re a lot more knowledgeable than me on slang internals so I think it
would be much quicker for you to find a fix.

For the bigger picture I want to keep the scope of this PR small—only
adding a few autocompletions for common syntax as well as creating the
framework for defining when a auto completion should be suggested.

Later prs can flesh out the snippets.

Best
-Evan

evanwporter:
I don’t think I’ll I have enough time to finish this in the future :(

PR #350, opened 2026-05-22
semantic tokens

### LINKED ISSUE #268: Implement Semantic Tokens

evanwporter:
Closes #268 

Follows pretty closely what was discussed in #268.

Here is my progress so far (from a generic part of my code for another project).

image

Still need to do the delta also lots of clean up, also it should recognize trivia.


evanwporter:
image

Also at least `word_t` is highlighted right now. Either way once this gets merged we'll have a much better framework to solve these issues.

evanwporter:
image

little buggy but it works somewhat

PR #352, opened 2026-05-23
[Neovim] Semantic tokens -- inactive regions

evanwporter:
Semantic tokens (#350) is a lot more work than I'm willing to put in (for now!), but we can still leverage them to provide inactive region support to editors that don't easily support `textDocument/inactiveRegions` (ie: neovim). This is what clangd does see

https://clangd.llvm.org/features#kinds

NEOVIM for reference:

image

In a future PR I'll add the ability to disable certain semantic tokens from appearing ala:

https://clangd.llvm.org/config.html#semantic-tokens

### DIFF
--- a/clients/vscode/resources/config.schema.json
+++ b/clients/vscode/resources/config.schema.json
@@ -100,6 +100,10 @@
           "items": {
             "$ref": "#/$defs/Config__Build"
           }
+        },
+        "semanticTokens": {
+          "$ref": "#/$defs/Config__SemanticTokensConfig",
+          "description": "Configure semantic highlighting."
         }
       },
       "required": []
@@ -201,6 +205,26 @@
         "macroArgNames": {
           "type": "integer",
           "description": "Macro argument hints: 0=off, N=only calls with >=N args"
+        }
+      },
+      "required": []
+    },
+    "Config__SemanticTokensConfig": {
+      "type": "object",
+      "properties": {
+        "enabled": {
+          "description": "Whether semantic tokens should be provided by the server.",
+          "type": "boolean"
+        },
+        "disabledKinds": {
+          "type": "array",
+          "description": "Specify semantic token kinds that slang-server should not send to client.",
+          "items": {
+            "type": "string",
+            "enum": [
+              "InactiveCode"
+            ]
+          }
         }
       },
       "required": []
--- a/clients/vscode/src/config.gen.ts
+++ b/clients/vscode/src/config.gen.ts
@@ -38,6 +38,8 @@ export interface Config {
   inlayHints?: Config__InlayHints
   /** Builds for direct .f selection or command-based .f generation */
   builds?: Config__Build[]
+  /** Configure semantic highlighting. */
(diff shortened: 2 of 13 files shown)


evanwporter:
~~Should be good to go, just need to add doc comments....~~

evanwporter:
@AndrewNolte gentle ping. You can test this out in vscode by flipping the inactiveRegionsSupported bool.

AndrewNolte:
I thought we were going with the client handler for Neovim semantic tokens?

evanwporter:
Yeah, but some people may not be using the neovim plugin, or they may be
using an entirely different editor like vim, emacs, zed, helix, etc…

This way they still get access to the feature.

AndrewNolte:
I think regardless we'd want semantic tokens to be configurable, as well as this ability for disabled regions. 

I feel like people would still want their syntax highlights in disabled sections? Especially bc in SV these could be massive regions of a file.

evanwporter:
Clangd allows for certain semantic tokens to be disabled. That would be the
next pr after this is merged—or it could just be added to this directly.

AndrewNolte:
Let's go ahead and make a semantic tokens config in this pr, with (enabled, disabledTokens) options, with both defaulted to false for now.

sgaulter:
If we're doing this, I don't think we should merge #402 as well. This is functionally cleaner than manually fiddling around with Neovim extmarks in the client.

Unless I've missed anything; does #402 provide anything that this doesn't?

AndrewNolte:
Handling it client side should allow for still showing syntax highlighting on disabled regions, the text will just be faded out a bit

evanwporter:
@AndrewNolte ready for review

evanwporter:
There's an optimization here if we use multiline semantic tokens (`multilineTokenSupport`). Idk how many editors support this though (vscode for one does not)

sgaulter:
> Handling it client side should allow for still showing syntax highlighting on disabled regions, the text will just be faded out a bit

Neovim unfortunately has no "opacity" parameter for text like VSCode, so that would be non-trivial. You'd have to walk the ranges, look at the existing highlights, calculate a blended colour and reapply a new highlight extmark. While resolving races with Treesitter, which is also updating highlight extmarks as the source changes.

If the first pass for Neovim is just greyed out comment-style blocks for inactive code, then this PR gets that without any client side changes, superseding #402. If we went on to implement "full" semantic tokens later, we could get the washed-out syntax in inactive regions with an "inactive" tokenModifier and some minimal client-side work.

evanwporter:
The other reason to handle it client side is because it’s more efficient
since you can process entire blocks of text. However it seems like neovim
supports multi line tokens so this isn’t that big of a deal:

https://github.com/neovim/neovim/blob/bc3757b7bf9326a5856f3ead844f71dc59dea206/runtime/lua/vim/lsp/protocol.lua#L420

> If we went on to implement "full" semantic tokens later

I tried doing this in #350. I found it’s a lot of work while being
significantly slower than tree sitter. :(

evanwporter:
@AndrewNolte lgtm!

PR #364, opened 2026-05-29
Append `TOP.` to top level module waveform path when top level has ports

evanwporter:
(Fixes verilator specific quirk)

I'm not sure if this is the best way to handle this since it could conflict with how other simulators draw up waveforms, but this is a very real problem I encounter quite a bit, since almost all of my testing is done in verilator.

### DIFF
--- a/clients/vscode/src/sidebar/ProjectComponent.ts
+++ b/clients/vscode/src/sidebar/ProjectComponent.ts
@@ -1042,7 +1042,6 @@ export class ProjectComponent
         if (didToggle) {
           void this.reveal(item)
         }
-
         await item.showChildrenInWaveform(this.logger)
       }
     }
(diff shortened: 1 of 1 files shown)


AndrewNolte:
Yeah I'd say this is better addressed on the verilator side

evanwporter:
> Yeah I'd say this is better addressed on the verilator side

Alright I think I'll just create a fork of the slang-server extension and use that. I find this pretty important, but I understand why it can't go upstream.

AndrewNolte:
There's this- https://verilator.org/guide/latest/exe_verilator.html#cmdoption-no-trace-top

evanwporter:
Oh this is exactly what I needed. Tysm

evanwporter:
> There's this- https://verilator.org/guide/latest/exe_verilator.html#cmdoption-no-trace-top

Unfortunately this was not a fix :(

I'm going to make an issue on the verilator page

PR #365, opened 2026-05-30
Expand location-only diagnostic ranges over identifiers

### LINKED ISSUE #197: Only first symbol highlighted for redefined variable
Image

Also, it marked as unused for second declaration (maybe its ok)
Image

jeet-dekivadia:
## Summary
- expand diagnostics that only carry a source location to the identifier token at that location
- apply the same range handling to related note locations after mapping note ranges
- add regression coverage for redefinition diagnostics so the duplicate identifier range covers the whole name

Fixes #197.

## Correctness
Location-only diagnostics arrive as a zero-width range. The helper looks up the syntax token at that offset and expands the end position only when the token is an identifier; if lookup fails or the token is not an identifier, it preserves the original range. Existing non-empty diagnostic ranges remain unchanged.

## Remote Linux validation
- `git diff --check`
- `uv venv`
- `source .venv/bin/activate`
- `uv sync`

### DIFF
--- a/src/ServerDiagClient.cpp
+++ b/src/ServerDiagClient.cpp
@@ -26,6 +26,60 @@
 #include "slang/text/SourceManager.h"
 namespace server {
 
+bool isSimpleIdentifierChar(char c) {
+    return (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9') ||
+           c == '_' || c == '$';
+}
+
+bool isTokenWhitespace(char c) {
+    return c == '\0' || c == ' ' || c == '\t' || c == '\n' || c == '\r' || c == '\f' ||
+           c == '\v';
+}
+
+lsp::Location toDiagnosticLocation(SourceLocation loc, const SourceManager& sourceManager) {
+    if (!sourceManager.isFileLoc(loc)) {
+        return toLocation(loc, sourceManager);
+    }
+
+    auto text = sourceManager.getSourceText(loc.buffer());
+    auto offset = loc.offset();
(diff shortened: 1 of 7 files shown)


src/ServerDiagClient.cpp
```diff
@@ -26,6 +26,60 @@
 #include "slang/text/SourceManager.h"
 namespace server {
 
+bool isSimpleIdentifierChar(char c) {
```
AndrewNolte, line 29: Similar functions already exist in `external/slang/include/slang/text/CharInfo.h`

AndrewNolte:
I mentioned this on the issue- this is better patched on slang, and I don't think it makes sense to try to retokenize on the server side. Slang should be patched to highlight the token range.

jeet-dekivadia:
Thanks for the direction. Agreed: this should be fixed in Slang so diagnostic ranges are correct at the source rather than retokenized in the server. I will leave the server-side patch closed.

jeet-dekivadia:
Thanks again. I traced this into Slang's `Scope::reportNameConflict`: the redefinition diagnostic and its previous-definition note are both emitted from `Symbol::location`, even though `Scope::addDiag` and `Diagnostic::addNote` already accept `SourceRange`. `slang-server` currently pins `AndrewNolte/slang@69cbd702f641f237372af5e5918d921b9941e138`, and I do not see the range fix in canonical Slang `master` or the current fork branches.

I have a small library-side patch with a regression asserting that both spans cover the complete identifier. Since `slang-server` consumes Andrew's fork, would you prefer the follow-up against `MikePopoloski/slang` or `AndrewNolte/slang`?

PR #366, opened 2026-05-30
Distinguish net and register hovers

### LINKED ISSUE #351: distinguish net vs register in hovers
It'd be helpful to distinguish between nets and registers in the hovers.

By net I mean a wire or a logic that is combinationally assigned.

By register I mean a logic that is assigned via a latch / FF

jeet-dekivadia:
## Summary
- reuse shallow driver analysis for hover classification instead of discarding it after diagnostics
- label continuously driven and `always_comb` variables as `Net`
- label `always_ff` and `always_latch` variables as `Register`
- retain the existing `Variable` label when the driver semantics are not definitive

## Validation
- `git diff --check HEAD^ HEAD`
- `/Library/Developer/CommandLineTools/usr/bin/clang-format -i include/Hovers.h include/document/ShallowAnalysis.h src/Hovers.cpp src/ServerDriver.cpp src/document/ShallowAnalysis.cpp tests/cpp/HoverTests.cpp`
- added hover regression coverage for explicit wires, continuous assignments, `always_comb`, `always_ff`, `always_latch`, and undriven variables

Fixes #351

## Remote Linux validation
- `git diff --check`
- `uv venv`
- `source .venv/bin/activate`
- `uv sync`

### DIFF
--- a/include/Hovers.h
+++ b/include/Hovers.h
@@ -10,6 +10,8 @@
 namespace server {
 
 lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
-                            const DefinitionInfo& info, const Config::HoverConfig& hovers);
+                            const DefinitionInfo& info,
+                            std::optional<std::string_view> variableKind,
+                            const Config::HoverConfig& hovers);
 
 }
--- a/include/document/ShallowAnalysis.h
+++ b/include/document/ShallowAnalysis.h
@@ -31,6 +31,11 @@
 #include "slang/text/SourceLocation.h"
 #include "slang/text/SourceManager.h"
 #include "slang/util/Bag.h"
+
+namespace slang::analysis {
+class AnalysisManager;
+}
+
 namespace server {
 using namespace slang;
 struct DefinitionInfo {
@@ -71,6 +76,7 @@ public:
     ShallowAnalysis(SourceManager& sourceManager, slang::BufferID buffer,
                     std::shared_ptr<slang::syntax::SyntaxTree> tree, slang::Bag options,
                     const std::vector<std::shared_ptr<slang::syntax::SyntaxTree>>& allTrees = {});
+    ~ShallowAnalysis();
 
     /// @brief Retrieves document symbols for LSP outline view, called right after open
     /// @return Vector of LSP document symbols representing the document structure
@@ -165,6 +171,9 @@ public:
     /// @return The analysis diagnostics (owned by internal AnalysisManager)
     Diagnostics getAnalysisDiags();
 
+    /// @brief Returns a more specific kind for variables with definitive driver semantics
+    std::optional<std::string_view> getVariableKind(const slang::ast::Symbol& symbol);
+
 private:
     /// Reference to the source manager. Not const because we may need to parse macro args.
     SourceManager& m_sourceManager;
@@ -183,6 +192,9 @@ private:
 
     /// Analysis options for driver analysis (numThreads=1 to avoid persistent threads)
     slang::analysis::AnalysisOptions m_analysisOptions;
+
+    /// Driver analysis is shared by diagnostics and hover classification.
+    std::unique_ptr<slang::analysis::AnalysisManager> m_driverAnalysis;
 
     /// Symbol tree visitor for /documentSymbols
     /// Currently this is relies on syntax, but we should switch it to use the shallow compilation
@@ -218,6 +230,8 @@ private:
     /// @param node The syntax node to search from
     /// @return Pointer to the name syntax, or nullptr if none found
     const slang::syntax::NameSyntax* findNameSyntax(const slang::syntax::SyntaxNode& node) const;
+
+    slang::analysis::AnalysisManager& getDriverAnalysis();
 };
 
 } // namespace server
--- a/src/Hovers.cpp
+++ b/src/Hovers.cpp
@@ -21,7 +21,9 @@
 namespace server {
 
 lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
-                            const DefinitionInfo& info, const Config::HoverConfig& hovers) {
+                            const DefinitionInfo& info,
+                            std::optional<std::string_view> variableKind,
+                            const Config::HoverConfig& hovers) {
     markup::Document doc;
 
     auto& infoPg = doc.addParagraph();
@@ -29,7 +31,8 @@ lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
     if (info.symbol) {
         // <Kind/Type> <Name> in <Scope>
 
-        infoPg.appendBold(toString(info.symbol->kind)).appendCode(info.symbol->name);
+        infoPg.appendBold(variableKind.value_or(toString(info.symbol->kind)))
+            .appendCode(info.symbol->name);
 
         auto symbolScope = info.symbol->getParentScope();
         auto& parentSym = symbolScope->asSymbol();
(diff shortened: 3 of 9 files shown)


toddstrader:
Thanks for the PR.

Sorry to be nit picky here, but I'd prefer that Slang Server not be the one to canonize terminology.  I think it would be better if the Slang library had a method which would return `Register`, `Net`, etc. when asked about these things.  Also, `Register` does not seem right to me for `always_latch`.  Yes, it's sequential, but a latch is different than a register.  But again, I think that discussion is better conducted in a PR on the library.

I think once the terminology is settled and enshrined upstream, the PR makes sense.

toddstrader:
Also, is the intent that `always` driven things remain `Variable`?  If so, I think that's a reasonable trade off since deducing / synthesizing the driver type for those blocks is pretty hairy.  But I'd suggest a comment to this effect in `getVariableKind()` (wherever it ends up living) so that we know it was intentionally skipped and not just forgotten.

jeet-dekivadia:
Thanks, that makes sense. I agree the classification and terminology should live in Slang, including treating latches separately and documenting the intentionally conservative behavior for plain `always` blocks. I am closing this server-side patch rather than establishing those semantics here.

evanwporter:
> Also, is the intent that always driven things remain Variable? If so, I think that's a reasonable trade off since deducing / synthesizing the driver type for those blocks is pretty hairy. But I'd suggest a comment to this effect in getVariableKind() (wherever it ends up living) so that we know it was intentionally skipped and not just forgotten.

Hey Todd, 

I created the linked issue because I think it would be nice to differentiate between the different ways a logic can be assigned.  Also, I didn't know that a register and a latch were different so thanks for the clarification.

When I see a `wire`/`net` in the hover, I understand that it represents an actual net connection:

image

But when I see `logic`/`variable`, it is less clear to me what role the signal is playing in the design. `logic` can be used in several contexts, so the hover does not immediately tell me whether I should think of it as combinational logic, a register, a latch, etc.

image

Code from screenshots:

```
module Test (
    input logic clk,
    input logic a_in
);
  logic a;
  assign a = a_in;

  wire b;
  assign b = a_in;

  always_comb begin
    if (a) begin
      $display("lala");
    end

    if (b) begin
      $display("yaya");
    end
  end
endmodule : Test
```

That might not answer your question, but maybe it helps clarify the original issue.

Jeet, thanks for taking a stab at closing the issue.

jeet-dekivadia:
Thanks Evan, really appreciate you saying that.

I also learned a lot from Todd’s feedback here, especially the point that the hover terminology should be owned upstream by Slang rather than inferred and canonized inside the server. That makes sense to me, so I’ll leave this closed rather than stretching this patch beyond the right architectural boundary.

Separately, I took a pass at a couple of other issues that looked self-contained and useful:

- [#371](https://github.com/hudson-trading/slang-server/pull/371): fixes the GCC 15 expand-macro initialization build failure, and keeps the GNU warning enforcement scoped to server code.
- [#367](https://github.com/hudson-trading/slang-server/pull/367): fixes package dependencies flowing through interfaces, with a regression fixture for an interface-imported payload type.

Whenever someone has bandwidth, I’d really appreciate a look at those too. I tried to keep both changes narrow, include regression coverage / validation, and stay aligned with the existing server behavior. Very happy to adjust quickly if either direction is not quite what you want.

Thanks again for the thoughtful guidance here, and for maintaining this project. It has been genuinely fun to work through these issues.

toddstrader:
I wish this were simple.  But it is not.

First off, if either of you don't already have it, it's probably a good idea to download the SystemVerilog LRM (IEEE 1800):
https://ieeexplore.ieee.org/document/10458102
It looks like it's paywalled, but if you sign up for IEEE GET (yeah, the account / program is called GET) you can download it for free.  Or you can just google for it.

I'd suggest checking out **6.5 Nets and variables** to start.  But yeah, we'll want to be careful about nomenclature because it's easy to solve these problems for the "reasonable" set of things.  But many people have a different definition of reasonable.

For example, you are correct that a `wire` is strictly a connection between a driver and one or more loads.  And I would also probably also use the terms `wire` and `net` interchangeably based on my usual use cases.  But if you look at **6.6 Net types** you'll find that there are many more net types than just `wire`.  And I'm sure that some people who will ultimately use Slang Server will be concerned with, e.g. `tri`.

The word "register" is unfortunately overloaded.  From my experience when hardware people say "register" they mean a flip flop, like a DQ flip flop.  However, `reg` in SystemVerilog is one of a few variable types (see **6.8 Variable declarations**).  It can be used to create a flip flop, but is not necessarily one.  SystemVerilog variables can be procedurally assigned, whereas nets cannot.  And you are correct that this property is not enough to determine what the variable "is".  E.g. is it a register / flip flop, latch or does it act as a net?  We need to look at the driver to figure out what the `reg` or `logic` "is".  `always_ff`, `always_latch` and `always_comb` definitely help because the express intent and tools are supposed to enforce it.  `always` is a problem because you basically need to synthesize the logic to try to figure out what the variable "is".

Also, someone realized the problem with the `reg` keyword at some point (**6.11.2 2-state (two-value) and 4-state (four-value) data types**):
`The keyword reg does not always accurately describe user intent, as it could be perceived to imply a hardware register. The keyword logic is a more descriptive term. logic and reg denote the same type.`

My suggestion here is to stick as closely to the SystemVerilog nomenclature as possible.  I believe I know what you're after, but I think it's more than a little hairy when you get into it.  It might be more verbose than you were hoping for, but one thing I'd suggest is that you can label things as:
`net|variable (always|always_comb|alawys_ff|always_latch)`
or something to specify the umbrella data type and the process type which drives it.  Actually, it's more complicated than that because `always` driven things can be driven by multiple procs (including `initial`).  So even that's not a solid answer.  And you could collapse things like `variable: always_ff` to `always_ff` or just `ff` or `flip-flop` or something.  The problem is that `always_ff` is itself misleading because `always_ff` just means that you can't have blocking assignments and you can use it to model other things like memories.  I guess `always_no_blocking` didn't roll off the tongue as well.

evanwporter:
Hey Todd,

Thanks for the info and the reading recs. From the standard I learned that you could do something like this: 

```
wire struct structA
```

So, I realized one solution to #351 could be:

image

(or alternatively you can swap the struct for a logic for the same effect).

Relevant code:

```
  typedef struct {logic a;} s0_t;

  s0_t s0_r;
  wire s0_t s0 = s0_r;

  always_comb begin
    s0_r.a = 1;
  end

  always_comb begin
    if (s0.a) begin
      $display("wow");
    end
  end
```

It's definitely a filthy solution, so I don't think I'll ever use it lol.

> It might be more verbose than you were hoping for, but one thing I'd suggest is that you can label things as:
net|variable (always|always_comb|alawys_ff|always_latch) or something to specify the umbrella data type and the process type which drives it. 

I think this is the best solution. I raised a related issue in the slang repo after your first comment, and the maintainer suggested using:

https://github.com/MikePopoloski/slang/blob/7b458f1361350a66a400fa0571e7a73d0823a4bf/include/slang/analysis/DriverTracker.h#L95-L96

https://github.com/MikePopoloski/slang/blob/7b458f1361350a66a400fa0571e7a73d0823a4bf/include/slang/analysis/ValueDriver.h#L31-L43

When `getDrivers` returns an array larger than 1, the hovers could display nothing, or it could display `mixed`.

> Actually, it's more complicated than that because always driven things can be driven by multiple procs (including initial).

I didn't realize this. 

You see from section 9.2 of the standard, it says of `always_comb` and `always_ff`; `The variables written on the left-hand side of assignments shall not be written to by any other process`. 

But the same rule doesn't apply for `always`. So, is this just legacy baggage from verilog, or is there a more nuanced answer here?

evanwporter:
See #388

PR #367, opened 2026-05-30
Load package dependencies of interfaces

### LINKED ISSUE #349: Incomplete type when I don't import the package
Consider the following three files:

```
module producer (
    simple_if.producer_mp bus
);
  always_comb begin
    bus.payload.valid = 1'b1;
    bus.payload.data  = 8'h42;
  end
endmodule


module consumer (
    simple_if.consumer_mp bus
);
  always_comb begin
    if (bus.payload.valid) begin
      // use bus.payload.data
    end
  end
endmodule


module top;
  simple_if bus ();

  producer u_producer (.bus(bus));

  consumer u_consumer (.bus(bus));
endmodule
```

```
package zhe_pkg;
  typedef struct packed {
    logic valid;
    logic data;
  } payload_t;

endpackage
```

```
import zhe_pkg::*;

interface simple_if;

  payload_t payload;

  modport producer_mp(output payload);

  modport consumer_mp(input payload);
endinterface
```

Notice how the `payload_t` is only imported into the interface file. Thus, when I hover over the one of the module declarations I get `incomplete type` see:

Image

This is solved when `payload_t` is imported into the first file above. However importing it gives unused import warning:

Image

Though it does work.

Image

This is somewhat self-contradictory and the incomplete type can be confusing and unhelpful.

A good fix would be to pull in the implied import of `payload_t` in files where it's not explicitly imported but still used.

This may be an `slang` issue.

jeet-dekivadia:
## Summary
- queue newly discovered package and interface trees for recursive dependency discovery
- add interface-imported packages to shallow compilation without recursively walking discovered modules
- add a regression fixture where a module depends on an interface that imports its payload type from a package

## Correctness
The originating tree is always inspected. Discovered documents are always added to the shallow dependency set, but only package and interface syntax trees are queued for another discovery step. For the regression graph, the walk reaches `interface_consumer -> transitive_if -> transitive_pkg`. A module discovered beneath another file is not queued, so the loader does not recursively follow the design hierarchy.

Fixes #349

## Remote Linux validation
- `git diff --check`
- `uv run pre-commit run --all-files`
- `uv venv`
- `source .venv/bin/activate`
- `uv sync`

### DIFF
--- a/src/ServerDriver.cpp
+++ b/src/ServerDriver.cpp
@@ -349,13 +349,8 @@ std::vector<std::shared_ptr<SlangDoc>> ServerDriver::getDependentDocs(
                 result.push_back(newdoc);
                 docs[newdoc->getURI()] = newdoc;
 
-                // Only add packages to the queue for recursive processing
-                for (auto& [decl, _] : newdoc->getSyntaxTree()->getMetadata().nodeMeta) {
-                    if (decl->kind == syntax::SyntaxKind::PackageDeclaration) {
-                        treesToProcess.push(newdoc->getSyntaxTree());
-                        break;
-                    }
-                }
+                // Interfaces and modules can add package dependencies of their own.
+                treesToProcess.push(newdoc->getSyntaxTree());
             }
             else {
                 ERROR("No doc found for {}", filePath);
--- a/src/ServerDriver.cpp
+++ b/src/ServerDriver.cpp
@@ -349,8 +349,14 @@ std::vector<std::shared_ptr<SlangDoc>> ServerDriver::getDependentDocs(
                 result.push_back(newdoc);
                 docs[newdoc->getURI()] = newdoc;
 
-                // Interfaces and modules can add package dependencies of their own.
-                treesToProcess.push(newdoc->getSyntaxTree());
+                // Packages and interfaces can add package dependencies of their own.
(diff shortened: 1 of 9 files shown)


src/ServerDriver.cpp
```diff
@@ -349,13 +349,8 @@ std::vector<std::shared_ptr<SlangDoc>> ServerDriver::getDependentDocs(
...
 
-                // Only add packages to the queue for recursive processing
-                for (auto& [decl, _] : newdoc->getSyntaxTree()->getMetadata().nodeMeta) {
-                    if (decl->kind == syntax::SyntaxKind::PackageDeclaration) {
-                        treesToProcess.push(newdoc->getSyntaxTree());
-                        break;
-                    }
-                }
+                // Interfaces and modules can add package dependencies of their own.
+                treesToProcess.push(newdoc->getSyntaxTree());
```
AndrewNolte: This will result in a whole design being loaded in shallow compilation; we don't want this.

jeet-dekivadia: Thanks for catching this. I narrowed the queue so discovered packages and interfaces are inspected recursively, while discovered modules are still added as shallow dependencies but are not queued. That keeps the interface-imported package case working without following the design hierarchy. I reran the Linux configure, build, CTest, and system-test lane successfully.

AndrewNolte:
Implemented in https://github.com/hudson-trading/slang-server/pull/404

PR #378, opened 2026-06-06
Vscode auto-install: Select old-linux binary on systems with GLIBC < 2.39

ya-uhs:
Addresses https://github.com/hudson-trading/slang-server/issues/384

## Problem

On RHEL 8 / AlmaLinux 8 (GLIBC 2.28), the VSCode extension fails to start slang-server v0.2.6. The extension downloads `slang-server-linux-x64-gcc.tar.gz`, which was built on Ubuntu 24.04 and requires GLIBC 2.39 — unavailable on RHEL 8 family systems.

The v0.2.6 release already ships `slang-server-old-linux-x64-gcc.tar.gz` (built on Rocky Linux 8, requires GLIBC ≥ 2.28), but the extension never selects it.

## Fix

Add `parseGlibcVersion()` and `isOldGlibc()` to `platform.ts`. On Linux, `isOldGlibc()` calls `ldd --version` and returns `true` if GLIBC < 2.39. `extension.ts` uses this to select the correct asset at install time.

```ts
linux: isOldGlibc()
  ? 'slang-server-old-linux-x64-gcc.tar.gz'
  : 'slang-server-linux-x64-gcc.tar.gz',
```

## Testing

Ran unit tests and verified `isOldGlibc()` against a live AlmaLinux 8 container (`ldd (GNU libc) 2.28` → `true`).

### DIFF
--- a/clients/vscode/src/extension.ts
+++ b/clients/vscode/src/extension.ts
@@ -21,6 +21,7 @@ import {
   ViewContainerSpec,
 } from './lib/libconfig'
 import { PathConfigObject } from './lib/pathConfig'
+import { isOldGlibc } from './lib/platform'
 import { ProjectComponent } from './sidebar/ProjectComponent'
 import * as slang from './SlangInterface'
 import {
@@ -143,7 +144,9 @@ File input is sent to stdin, and formatted output is read from stdout.',
         githubRepo: 'hudson-trading/slang-server',
         assetNames: {
           windows: 'slang-server-windows-x64.zip',
-          linux: 'slang-server-linux-x64-gcc.tar.gz',
+          linux: isOldGlibc()
+            ? 'slang-server-old-linux-x64-gcc.tar.gz'
+            : 'slang-server-linux-x64-gcc.tar.gz',
           mac: 'slang-server-macos.tar.gz',
         },
       },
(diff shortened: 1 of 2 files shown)


evanwporter:
Hey, I helped add the vscode installer feat, so I'm glad you're using it (or would use it after this PR goes through lol). One thing is, I think a more elegant solution than parsing the output of `ldd` is to use the node diagnostic report: 

https://nodejs.org/dist/latest-v16.x/docs/api/report.html

```
const report = process.report.getReport();
// parse glibc version
```

Note I found this feat here:

https://github.com/nodejs/node/issues/39877#issuecomment-905383436

ya-uhs:
Thanks for the suggestion! I've updated the implementation to use process.report.getReport() instead.

evanwporter:
Looks good two last nit picks:

(1) scrap the test. Typescript ensures it returns a bool so we don’t need
to test that lol.

(2) use the semver package to parse the versioning instead of doing it by
hand; ie:

semver.gt("1.2.3", "9.8.7"); // false

https://docs.npmjs.com/cli/v6/using-npm/semver

ya-uhs:
Done! Removed the test and switched to semver for version comparison.

AndrewNolte:
Thanks for the PR! We decided to go with these instead though, so closing. 
https://github.com/hudson-trading/slang-server/pull/385
https://github.com/hudson-trading/slang-server/pull/389

Lmk if there's a particular reason you think we should try to inspect the user's libc rather than just using the more compatible one.

PR #379, opened 2026-06-07
Use source buffer for explicit Neovim commands

### LINKED ISSUE #377: [neovim] Window focus not registering command
When I am clicked in this area, 

Image

and I run any `SlangSever XXXX` command I get this error:

Image

```
slang-server: no slang-server LSP client attached. Please upgrade slang-server and possibly also this plugin.
```

The solution is to switch back over here:

Image

This is fine I guess but intuitively I'd expect to be able to run commands anywhere in neovim.

I'm still learning vim, so while I have some semblance of an idea about why this is happening--the slang-server disappears when the buffer is no longer in focus--I'm not really sure what the best fix is.

jeet-dekivadia:
## Summary

- resolve explicit Neovim commands through the current buffer when it has the slang-server client, otherwise fall back to the most recently used visible SystemVerilog/Verilog source buffer
- keep cursor/current-file-sensitive commands tied to the active source buffer instead of guessing stale context
- replace the misleading upgrade hint for those context-sensitive panel invocations with a direct context error
- add Neovim regressions for both the explicit-command fallback and the context-sensitive refusal path

## Why

The Neovim command is global, but commands like `setBuildFile`, `openWaveform`, and `setTopLevel <file>` only need an attached slang-server client to send the request. When focus is inside a plugin panel, the panel buffer does not have the client attached, so those commands used to fail with `no slang-server LSP client attached` even though the source buffer is still open and attached.

Commands that depend on the current file or cursor position should not silently reuse an old source window. This keeps `addToWaves` and no-argument `setTopLevel` scoped to the current source buffer and gives a clearer message when run from the hierarchy panel.

Fixes #377

## Validation

- `git diff --check`
- `.venv/bin/pre-commit run --files clients/neovim/lua/slang-server/_lsp/capabilities.lua clients/neovim/lua/slang-server/_commands/addToWaves.lua clients/neovim/lua/slang-server/_commands/hierarchy.lua clients/neovim/lua/slang-server/_commands/openWaveform.lua clients/neovim/lua/slang-server/_commands/setBuildFile.lua clients/neovim/lua/slang-server/_commands/setTopLevel.lua clients/neovim/spec/plugin_spec.lua`
- `.venv/bin/python scripts/ci/neovim-sync.py --event pull_request --dry-run`

### DIFF
--- a/clients/neovim/lua/slang-server/_lsp/capabilities.lua
+++ b/clients/neovim/lua/slang-server/_lsp/capabilities.lua
@@ -3,6 +3,7 @@ local M = {}
 M.MIN_SERVER_VERSION = "0.2.2"
 
 local UPGRADE_HINT = "Please upgrade slang-server and possibly also this plugin."
+local SOURCE_FILETYPES = { "verilog", "systemverilog" }
 
 ---@param v string
 ---@return integer, integer, integer
@@ -36,6 +37,27 @@ function M.get_client(bufnr)
       end
    end
    return nil
+end
+
+---@return integer bufnr, integer? winid
+function M.get_source_context()
+   local bufnr = vim.api.nvim_get_current_buf()
+   if M.get_client(bufnr) then
+      return bufnr, vim.api.nvim_get_current_win()
+   end
+
+   local util = require("slang-server.util")
+   local source_win = util.last_win({ buflisted = true, filetype = SOURCE_FILETYPES })
+   if source_win and M.get_client(source_win.bufnr) then
+      return source_win.bufnr, source_win.winid
+   end
+
+   local source_buf = util.last_buf({ buflisted = true, filetype = SOURCE_FILETYPES })
+   if source_buf and M.get_client(source_buf.bufnr) then
+      return source_buf.bufnr, nil
+   end
+
+   return bufnr, nil
 end
 
 -- Per-client cached commands set. Keyed by client.id; populated lazily on the
--- a/clients/neovim/lua/slang-server/_lsp/capabilities.lua
+++ b/clients/neovim/lua/slang-server/_lsp/capabilities.lua
@@ -39,25 +39,25 @@ function M.get_client(bufnr)
    return nil
 end
 
----@return integer bufnr, integer? winid
+---@return integer bufnr
 function M.get_source_context()
    local bufnr = vim.api.nvim_get_current_buf()
    if M.get_client(bufnr) then
-      return bufnr, vim.api.nvim_get_current_win()
+      return bufnr
    end
 
    local util = require("slang-server.util")
    local source_win = util.last_win({ buflisted = true, filetype = SOURCE_FILETYPES })
    if source_win and M.get_client(source_win.bufnr) then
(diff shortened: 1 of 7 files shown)


toddstrader:
Agreed this is not ideal.  I am however a little uncertain of what the best solution is here.  Some commands like `openWaveform` and `setBuildFile`, you just want to work no matter what buffer you are in.  Something like `addToWaves` feels a bit more confusing.  If you've switched from SystemVerilog to Python an hour ago, do you really remember where you left your cursor for `addToWaves` to make sense without actually being in that buffer?  Same for `setTopLevel` if you aren't providing an argument.

At the very least you shouldn't get the notification to update the plugin.  I'd suggest only commands which don't require buffer context make sense to send when not actually in the buffer.  But I'm curious to hear other thoughts / usage scenarios.

Curious: @evanwporter what was the plugin command you expected to work in this context?

jeet-dekivadia:
Thanks, that distinction makes sense to me. I tightened the patch so only explicit commands fall back to the attached source buffer.

Concretely, `setBuildFile`, `openWaveform`, `hierarchy`, and `setTopLevel <file>` can still run from the hierarchy panel, but `addToWaves` and no-argument `setTopLevel` now require the active buffer to have the slang-server client attached. Those paths also return a direct context error instead of the upgrade hint. I updated the regression coverage for both cases and reran the local pre-commit + Neovim sync checks.

clients/neovim/lua/slang-server/_lsp/capabilities.lua
```diff
@@ -38,6 +39,27 @@ function M.get_client(bufnr)
    return nil
 end
 
+---@return integer bufnr, integer? winid
```
sgaulter: I don't see the returned `winid` being used at any call sites. Let's remove it if it's unnecessary.

jeet-dekivadia: Done in 77a5203. I removed the unused second return value from `get_source_context()` and kept the call sites as single-value consumers.

Validation:

```text
git diff --check
.venv/bin/pre-commit run --files clients/neovim/lua/slang-server/_lsp/capabilities.lua
.venv/bin/python scripts/ci/neovim-sync.py --event pull_request --dry-run
```

evanwporter:
> Agreed this is not ideal. I am however a little uncertain of what the best solution is here. Some commands like `openWaveform` and `setBuildFile`, you just want to work no matter what buffer you are in. Something like `addToWaves` feels a bit more confusing. If you've switched from SystemVerilog to Python an hour ago, do you really remember where you left your cursor for `addToWaves` to make sense without actually being in that buffer? Same for `setTopLevel` if you aren't providing an argument.
> 
> At the very least you shouldn't get the notification to update the plugin. I'd suggest only commands which don't require buffer context make sense to send when not actually in the buffer. But I'm curious to hear other thoughts / usage scenarios.
> 
> Curious: @evanwporter what was the plugin command you expected to work in this context?

I was trying to run `openWaveform`.

PR #386, opened 2026-06-10
cancel-in-progress on pull request pushes

### LINKED ISSUE #383: [ci] Add `concurrency` config to kill PR workflows when new commits are pushed
I make a PR. I push. CI starts. I fix a typo. I push. I get comments. I fix. Then I push. Then 25 minutes later the CI for the current commit starts.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-using-concurrency-and-the-default-behavior

ktbarrett:
Closes #383.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,6 +32,10 @@ on:
       - 'pyproject.toml'
       - 'uv.lock'
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
 jobs:
   build:
     name: ${{ matrix.name }}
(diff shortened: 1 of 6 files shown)


evanwporter:
@AndrewNolte can we revisit this PR? This is kinda annoying to have builds cancel midway through because you pushed a new commit or not be able to run a new workflow since an old one is running.

I don't really see the point since github foots the workflow bill.

ktbarrett:
> I don't really see the point since github foots the workflow bill.

Github has rate limits on Actions. Also why waste compute?

If you are pushing commits and expecting them all to complete you are liking doing something non-idiomatic. PRs are one change. You should be testing locally before pushing.

evanwporter:
Hey @ktbarrett, sorry I should’ve responded earlier. I had an issue with this PR when I was working on adding the musl builds git workflow. It wasn’t something I could really test locally (because it was a workflow), so I just opted for the strat of pushing a bunch of tiny commits and then look at the first commit to fail. Unfortunately this PR, was blocking me from doing so :(. But that pr has been merged and we’re good!

PR #388, opened 2026-06-10
Label `logic` driver source + kind in hovers

### LINKED ISSUE #351: distinguish net vs register in hovers

evanwporter:
Closes #351 

Copied from https://github.com/MikePopoloski/slang/issues/1861

Within SystemVerilog the `logic` keyword is somewhat unique in that its meaning isn't wholly contained within the declaration ("meaning" isn't quite the right word here but hopefully you get my point :) ).

Consider (FYI: I used chatgpt to generate the following examples):

### 1. `logic` assigned by a continuous assignment

```systemverilog
module ContinuousLogic (
    input logic a,
    output logic q
);
    assign q = a;
endmodule
```

### 2. `logic` assigned as combinational logic

```systemverilog
module CombinationalLogic (
    input logic a,
    input logic b,
    output logic q
);
    always_comb begin
        q = a & b;
    end
endmodule
```

### 3. `logic` assigned as a latch

```systemverilog
module LatchLogic (
    input logic en,
    input logic d,
    output logic q
);
    always_latch begin
        if (en)
            q = d;
    end
endmodule
```

```systemverilog
module InferredLatchLogic (
    input logic en,
    input logic d,
    output logic q
);
    always_comb begin
        if (en)
            q = d;
        // q is not assigned when en is false
    end
endmodule
```

### 4. `logic` assigned as a register / flip-flop

```systemverilog
module RegisterLogic (
    input logic clk,
    input logic d,
    output logic q
);
    always_ff @(posedge clk) begin
        q <= d;
    end
endmodule
```

In all 4 examples `q` is a `variable`, however the exact way that q is set changes in each example.

To this end this PR attemps to add an additional piece of info to the hovers that distinguishes between each of the different ways `q` can be set--ie: as a latch, net or register.

Images:

image


src/Hovers.cpp
```diff
@@ -54,16 +126,56 @@ lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
...
                 infoPg.appendText("Width: ")
                     .appendCode(fmt::format("{}", type.getBitWidth()))
                     .newLine();
             }
+
+            const auto drivers = analysis->getDrivers(valSym);
+            if (!drivers.empty()) {
+                /// We only show driver info for values that have a single unique driver.
+                /// `structs` and variables that use `always` statements can have multiple drivers.
+                std::vector<const slang::analysis::ValueDriver*> uniqueDrivers;
```
evanwporter: maybe hashing would be better here?

toddstrader:
This makes more sense to me re: terminology.  I'll leave it to @AndrewNolte to opine on the implementation.

include/document/ShallowAnalysis.h
```diff
@@ -181,6 +191,15 @@ class ShallowAnalysis {
     /// Compilation context for symbol resolution
     std::unique_ptr<slang::ast::Compilation> m_compilation;
 
+    /// Analysis manager for running driver analysis (multi-driven, unused, etc)
+    std::unique_ptr<slang::analysis::AnalysisManager> m_driverAnalysis;
```
AndrewNolte: You should be able to just check the unique ptr (instead of m_analysisRun)

src/Hovers.cpp
```diff
@@ -20,12 +21,93 @@
...
+
+    switch (source) {
+        case DriverSource::Initial:
+            return "initial";
+        case DriverSource::Final:
+            return "final";
+        case DriverSource::Always:
+            return "always";
+        case DriverSource::AlwaysComb:
+            return "always_comb";
```
AndrewNolte: There's `SemanticFacts::getProcedureKindStr` - we should upstream the commit that adds "other" though. We could still use that function here for now though if it qualifies as a ProceduralBlockKind

src/Hovers.cpp
```diff
@@ -20,12 +21,93 @@
...
+}
+
+std::string_view driverKindToString(const slang::analysis::DriverKind kind) {
+    using slang::analysis::DriverKind;
+
+    switch (kind) {
+        case DriverKind::Procedural:
+            return "procedural";
+        case DriverKind::Continuous:
+            return "continuous";
```
AndrewNolte: The toString for this enum should be fine

src/Hovers.cpp
evanwporter: subroutine too

src/Hovers.cpp
```diff
@@ -54,16 +136,57 @@ lsp::MarkupContent getHover(const SourceManager& sm, const BufferID docBuffer,
...
+
+                    const auto sameKindAndSource =
+                        [driver](const slang::analysis::ValueDriver* existing) {
+                            return existing && existing->kind == driver->kind &&
+                                   existing->source == driver->source;
+                        };
+
+                    if (std::ranges::find_if(uniqueDrivers, sameKindAndSource) ==
+                        uniqueDrivers.end()) {
+                        uniqueDrivers.push_back(driver);
```
AndrewNolte: We can short circuit if uniqueDrivers hits 2, rather than building up the list

evanwporter: What if in the future, we want to display multiple drivers? For example: `always`, `initial` and `final`

AndrewNolte:
Sorry did a big refactor here for hover logic- https://github.com/hudson-trading/slang-server/pull/339

evanwporter:
Nice fix. I like how it’s split up. It looks cleaner now.

Shouldn’t be too hard too resolve conflicts though it’s tough to say since
I’m on mobile

evanwporter:
it works but the way that the `extraDisplayNode` and the `ShallowAnalysis` is passed around could be improved

PR #396, opened 2026-06-16
Add driver info `assignment`

evanwporter:
Stacks onto #388 

Need to make the `extraDisplayNode` stuff a little cleaner.

This simply displays the assignment. ie:

```
logic hello;
assign hello = 1'b1
```

Images:

image


src/document/DefinitionInfo.cpp
```diff
@@ -35,15 +35,50 @@ using namespace slang;
...
+    for (auto cur = node; cur; cur = cur->parent) {
+        if (cur->kind == syntax::SyntaxKind::ContinuousAssign) {
+            return &selectDisplayNode(*cur);
+        }
+    }
+
+    return nullptr;
+}
+
+const syntax::SyntaxNode* renderSymbolHeader(markup::Paragraph& infoPg, const ast::Symbol& symbol,
```
AndrewNolte: This return is a bit awkward. Since it's only used once, let's inline it and split things out like showing the value/type into their own void functions. 

I think it also might be nicer to say `Driven continuously at:` and then the driver syntax, after the symbol syntax. However then it'd be weird if no syntax was rendered. I think for always_ff you could basically filter the statements for the drivers of that var + the control flow, since there's no dependencies in those statements. `always_comb` would be trickier - I think it's fine to just render the whole thing. Typically you'd just see `always_comb x = ... ` anyway. Actually I think filtering the `always_comb` statements would still be useful.

evanwporter: > I think it also might be nicer to say Driven continuously at: and then the driver syntax, after the symbol syntax. However then it'd be weird if no syntax was rendered. I think for always_ff you could basically filter the statements for the drivers of that var + the control flow, since there's no dependencies in those statements. always_comb would be trickier - I think it's fine to just render the whole thing.

I'm not following here? Are you suggesting we also render the variable w/i `always_ff`? 

That's certainty an option...but I think that's a little bit outside of the scope of what I was hoping to accomplish with this PR. 

You see the nice thing about having an `assign` is its usually only driven once and it can't have any conditional logic change the behavior. This makes it a good candidate for being displayed in the hovers.

>Typically you'd just see `always_comb x = ... ` anyway. 

That's interesting. I've always used the pattern:

```
logic x;
assign x = ...
```

Because I use this pattern so often I made this PR since I think it'd be helpful to display it.

src/document/DefinitionInfo.cpp
```diff
@@ -179,20 +207,42 @@ void renderMacroHeader(markup::Paragraph& infoPg, const DefinitionInfo::MacroTar
...
+    auto appendExtraDisplayNode = [&](std::string& code) {
+        if (extraDisplayNode) {
+            code += "\n";
+
+            const auto formattedCode = formatCode(*extraDisplayNode);
+
+            // Catches when people have long if/else (?/:) chains
+            // 300 is completely arbitrary and could probably be made into a config option or
+            // a compile time constant
+            if (formattedCode.size() <= 300)
```
evanwporter, line 200: Thoughts @AndrewNolte

I was thinking that for some large codebases there might be a large chain of ?/: which could clog up the hovers.

src/document/DefinitionInfo.cpp
evanwporter: > I think it also might be nicer to say Driven continuously at: and then the driver syntax, after the symbol syntax. However then it'd be weird if no syntax was rendered. I think for always_ff you could basically filter the statements for the drivers of that var + the control flow, since there's no dependencies in those statements. always_comb would be trickier - I think it's fine to just render the whole thing.

I'm not following here? Are you suggesting we also render the variable w/i `always_ff`? 

That's certainty an option...but I think that's a little bit outside of the scope of what I was hoping to accomplish with this PR. 

You see the nice thing about having an `assign` is its usually only driven once and it can't have any conditional logic change the behavior. This makes it a good candidate for being displayed in the hovers.

>Typically you'd just see `always_comb x = ... ` anyway. 

That's interesting. I've always used the pattern:

```
logic x;
assign x = ...
```

For this reason, I'd find this PR helpful.

tests/data/hdl_test.sv
```diff
@@ -121,13 +121,20 @@ module TestModule #(
...
+    assign thing = inst_array_test[0] ? 1'b1 : 1'b0;
+
     always_ff @(posedge clk) begin
         if (rst) begin
-            state <= test_pkg::STATE_A;
+            state   <= test_pkg::STATE_A;
             counter <= '0;
         end else begin
-            state <= state_next;
+            state   <= state_next;
```
AndrewNolte, line 136: Could we split out the test data changes into a separate commit/PR?

evanwporter:
Yeah I can do that —but I wrote the tests to test this pr

AndrewNolte:
It's nicer to see the diffs for test data change vs behavior change separately imo (for future prs)

AndrewNolte:
Of course it is a bit of a pain if you don't have stacked prs setup

src/document/DefinitionInfo.cpp
AndrewNolte: I don't think you addressed the main concern about renderSymbolHeader returning a syntax node pointer? The name of the function doesn't imply that it should be returning anything.

src/document/DefinitionInfo.cpp
AndrewNolte: I think it's fine to just render it always

evanwporter:
Cop out response but I have no idea where I was at with this pr

tests/data/hdl_test.sv
evanwporter: I guess...but isn't it better to keep them in the same PR to maintain the logical connection between them?

Like here's my changes and here are the accompanying tests. Instead of here are my changes and then like 5 commits later there's the tests for the pr

evanwporter:
Ayo ngl i used codex to handle the merge

evanwporter:
But I verified it works so :shrug:

AndrewNolte:
Going to be landing this as this stack- https://github.com/hudson-trading/slang-server/pull/442

AndrewNolte:
It moves the driver classification to the end of the hover btw, and shows the whole block

AndrewNolte:
Also provides a file link to the driver

PR #399, opened 2026-06-17
Neovim client config improvements

sgaulter:
- Remove the unintuitive `opts = {}` / `setup({})` requirement when using lazy.nvim.
- Document the manual installation & configuration methods, as well as the lazy.nvim ones.

### DIFF
--- a/clients/neovim/README.md
+++ b/clients/neovim/README.md
@@ -21,12 +21,56 @@ If installing with lazy.nvim, plugin dependencies are resolved automatically.
 * [nui.nvim](https://github.com/MunifTanjim/nui.nvim)
 
 ## Installation
-Use your favorite Neovim plugin manager to download and install the plugin.  If you happen to use lazy.nvim you can [install the plugin](https://www.lazyvim.org/configuration/plugins) by adding `~/.config/nvim/lua/plugins/slang-server.lua`:
+
+Clone and place the plugin directory in your Neovim runtimepath. It is lazily loaded by default on the first invocation of a `:SlangServer` command, so there's no need to rely on a plugin manager for lazy loading.
+
+Alternatively, use your favorite Neovim plugin manager to download and install the plugin.  If you happen to use lazy.nvim you can [install the plugin](https://www.lazyvim.org/configuration/plugins) by adding, e.g., `~/.config/nvim/lua/plugins/slang-server.lua`:
 ```lua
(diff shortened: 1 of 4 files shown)


clients/neovim/README.md
```diff
@@ -30,6 +33,47 @@ return {
...
+      SlangServerPackage = { fg = "#f48fb1" },
+      SlangServerPortInput = { fg = "#8bcd5b" },
+      SlangServerPortOutput = { fg = "#f65866" },
+      SlangServerPortInout = { fg = "#34bfd0" },
+      SlangServerParam = { fg = "#c75ae8" },
+      SlangServerLogic = { fg = "#dd9046" },
+      SlangServerReg = { fg = "#dd9046" },
+   },
+}
+```
```
toddstrader: Maybe link into the file instead of copying here?

clients/neovim/README.md
```diff
@@ -21,7 +21,10 @@ If installing with lazy.nvim, plugin dependencies are resolved automatically.
 * [nui.nvim](https://github.com/MunifTanjim/nui.nvim)
 
 ## Installation
-Use your favorite Neovim plugin manager to download and install the plugin.  If you happen to use lazy.nvim you can [install the plugin](https://www.lazyvim.org/configuration/plugins) by adding `~/.config/nvim/lua/plugins/slang-server.lua`:
+
+Clone and place the plugin directory in your Neovim runtimepath. It is lazily loaded by default on the first invocation of a `:SlangServer` command, so there's no need to rely on a plugin manager for lazy loading.
```
toddstrader: I guess I'd assume "use a plugin manager" is the default over "clone it yourself".  Thoughts?  I think whatever the common path is should be first.

clients/neovim/README.md
sgaulter: I very often look to a plugin's README.md for a quick configuration overview.

Is your concern visual noise, or potentially having to keep source and docs in sync? If it's the former, I could hide it in a `<detail>` block; if it's the latter, then that's more of a problem.

toddstrader: Both, but more the latter.  Can markdown pull in a range of lines in GH?  I have no idea.

sgaulter: It cannot 😞. No problem, I'll link to the config instead.

PR #402, opened 2026-06-18
[Neovim] Handle `textDocument/inactiveRegions`

evanwporter:
This adds the `inactiveRegions` capability (introduced in #287) to the neovim plugin.

We should still merge #352 since some people might not have the plugin, and some editors may not have support for this (ie: zed).

It works in basic cases, but there is an error when deleting many lines.

Feedback would be appreciated, since this is my first time doing substantial work w/ Neovim.

FYI: This PR is bootstrapped by AI.

Images:

image

### DIFF
--- a/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
+++ b/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
@@ -2,19 +2,18 @@ local M = {}
 
 local ns = vim.api.nvim_create_namespace("SlangServerInactiveRegions")
 
-local files = {}
-
 vim.api.nvim_set_hl(0, "SlangServerInactiveRegion", {
    link = "Comment",
    default = true,
 })
 
-local function apply_highlights(uri)
-   local ranges = files[uri]
-   if not ranges then
-      return
-   end
+---@class slang-server.InactiveRegionsParams
+---@field uri string
+---@field regions lsp.Range[]
 
+---@param uri string
+---@param ranges lsp.Range[]
+local function apply_highlights(uri, ranges)
    local bufnr = vim.fn.bufnr(vim.uri_to_fname(uri), false)
    if bufnr == -1 then
       return
@@ -32,6 +31,8 @@ local function apply_highlights(uri)
    end
 end
 
+---@param err lsp.ResponseError?
+---@param params slang-server.InactiveRegionsParams?
 function M.handler(err, params, _, _)
    if err then
       return
@@ -45,8 +46,7 @@ function M.handler(err, params, _, _)
       return
    end
 
-   files[params.uri] = params.regions
-   apply_highlights(params.uri)
+   apply_highlights(params.uri, params.regions)
 end
 
 return M
--- a/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
+++ b/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
@@ -1,5 +1,3 @@
--- lua/slang-server/_lsp/inactive_regions.lua
-
 local M = {}
 
 local ns = vim.api.nvim_create_namespace("SlangServerInactiveRegions")
--- a/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
+++ b/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
@@ -1,0 +1,54 @@
+-- lua/slang-server/_lsp/inactive_regions.lua
+
+local M = {}
+
+local ns = vim.api.nvim_create_namespace("SlangServerInactiveRegions")
+
+local files = {}
+
+vim.api.nvim_set_hl(0, "SlangServerInactiveRegion", {
+   link = "Comment",
+   default = true,
+})
+
+local function apply_highlights(uri)
+   local ranges = files[uri]
+   if not ranges then
+      return
+   end
+
+   local bufnr = vim.fn.bufnr(vim.uri_to_fname(uri), false)
+   if bufnr == -1 then
+      return
+   end
+
+   vim.api.nvim_buf_clear_namespace(bufnr, ns, 0, -1)
+
+   for _, range in ipairs(ranges) do
+      vim.api.nvim_buf_set_extmark(bufnr, ns, range.start.line, range.start.character, {
+         end_row = range["end"].line,
+         end_col = range["end"].character,
+         hl_group = "SlangServerInactiveRegion",
+         hl_eol = true,
+      })
+   end
+end
+
+function M.handler(err, params, _, _)
+   if err then
+      return
+   end
+
+   if not params then
+      return
+   end
+
+   if not params.uri or not params.regions then
+      return
+   end
+
+   files[params.uri] = params.regions
+   apply_highlights(params.uri)
+end
+
+return M
--- a/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
+++ b/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
@@ -7,10 +7,6 @@ vim.api.nvim_set_hl(0, "SlangServerInactiveRegion", {
    default = true,
 })
 
----@class slang-server.InactiveRegionsParams
----@field uri string
----@field regions lsp.Range[]
-
 ---@param uri string
 ---@param ranges lsp.Range[]
 local function apply_highlights(uri, ranges)
--- a/clients/neovim/lua/slang-server/init.lua
+++ b/clients/neovim/lua/slang-server/init.lua
@@ -13,7 +13,6 @@ M.setup = function(opts)
    config.update(opts)
 
    -- note: this config requires the slang_server to be explicitly named slang_server
-   -- so slang-server wouldn't work; idk the best practices here
    vim.lsp.config("slang_server", {
       capabilities = {
          experimental = {
--- a/clients/neovim/lua/slang-server/init.lua
+++ b/clients/neovim/lua/slang-server/init.lua
@@ -1,6 +1,7 @@
 -- Main module file
 
 local config = require("slang-server._core.config")
+local inactive_regions = require("slang-server._lsp.experimental.inactive_regions")
 
 ---@class SlangModule
 local M = {}
@@ -10,6 +11,22 @@ config.initialise()
 ---@param opts slang-server.config.Configuration?
 M.setup = function(opts)
    config.update(opts)
+
+   -- note: this config requires the slang_server to be explicitly named slang_server
+   -- so slang-server wouldn't work; idk the best practices here
+   vim.lsp.config("slang_server", {
+      capabilities = {
+         experimental = {
+            inactiveRegions = {
+               inactiveRegions = true,
+            },
+         },
+      },
+
+      handlers = {
+         ["textDocument/inactiveRegions"] = inactive_regions.handler,
+      },
+   })
 end
 
 return M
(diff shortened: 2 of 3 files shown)


clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
```diff
@@ -0,0 +1,48 @@
+local M = {}
+
+local ns = vim.api.nvim_create_namespace("SlangServerInactiveRegions")
+
+vim.api.nvim_set_hl(0, "SlangServerInactiveRegion", {
```
evanwporter, line 5: i thought about moving this to the highlight lua file, but I was having issues with the load order when I did

clients/neovim/lua/slang-server/init.lua
sgaulter, line 1: `init.lua` only gets executed when calling `require('slang-server').setup({})`, which may not ever happen, depending on the installation method (particularly after #399).

Prior to this, all plugin functionality has been lazy-loaded on the first `:SlangServer` invocation, but this is a new type of behaviour. The eager execution path when opening an SV buffer is `ftplugin/systemverilog.lua`, so the handler will have to be registered somewhere under there.

Also, while this `init.lua` is currently light, I'd prefer not to `require('slang-server')` under `ftplugin/systemverilog.lua` as it leaves a possible future footgun for more code to creep into the eager execution path.

I think the best approach is to move the whole handler attachment into `inactive_regions.lua` and `require('slang-server._lsp.experimental.inactive_regions')` in `ftplugin/systemeverilog.lua`.

clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
sgaulter: `highlights.lua` is for sure where this should live, with the value ideally pulled from `config.lua` like the other highlight groups.

I'm not sure what load order problems there could be if you `local highlights = require('slang-server._core.highlights')` above. If you want to share specifics of the errors hit then we can debug.

clients/neovim/lua/slang-server/init.lua
```diff
@@ -10,6 +11,21 @@ config.initialise()
...
+            inactiveRegions = {
+               inactiveRegions = true,
+            },
+         },
+      },
+
+      handlers = {
+         ["textDocument/inactiveRegions"] = inactive_regions.handler,
+      },
+   })
```
sgaulter, line 28: I think this races with the LSP initialisation. Once the `slang-server` LSP client has started, updating its `capabilities` retroactively won't have any effect, and we can't guarantee when this plugin will be loaded relative to the client.

Possible solutions:
- Document that the user needs to add this to their own LSP config if they want to use this feature.
- If the client is already running, update its capabilities and notify the server with `"workspace/didChangeConfiguration"` (if that's supported, @AndrewNolte?)

The former is obviously less than ideal. Thoughts, @toddstrader?

sgaulter: Incidentally the handler can be added after the client has started; it just has to be added to `local client = vim.lsp.get_clients("slang_server")` instead.

clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
evanwporter: Consider the following diff:

```
 ~/slang-server   neovim-inactive  git --no-pager diff                                                               Fri 19 Jun 2026 03:32:26 PM UTC
diff --git a/clients/neovim/lua/slang-server/_core/highlights.lua b/clients/neovim/lua/slang-server/_core/highlights.lua
index cba498c..28dc0bc 100644
--- a/clients/neovim/lua/slang-server/_core/highlights.lua
+++ b/clients/neovim/lua/slang-server/_core/highlights.lua
@@ -20,10 +20,17 @@ M.HIER_SUBTLE = "Comment"
 M.HIER_VALUE = "Constant"
 M.HIER_ERROR = "ErrorMsg"

+M.INACTIVE_REGIONS = "SlangServerInactiveRegion"
+
 for hl_name, hl_opts in pairs(config.highlights) do
    vim.api.nvim_set_hl(M.ns_id, hl_name, hl_opts)
 end

 vim.api.nvim_set_hl_ns(M.ns_id)

+vim.api.nvim_set_hl(0, M.INACTIVE_REGIONS, {
+   link = "Comment",
+   default = true,
+})
+
 return M
diff --git a/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua b/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
index bba6acf..d31e657 100644
--- a/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
+++ b/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
@@ -1,11 +1,8 @@
 local M = {}

-local ns = vim.api.nvim_create_namespace("SlangServerInactiveRegions")
+local hl = require("slang-server._core.highlights")

-vim.api.nvim_set_hl(0, "SlangServerInactiveRegion", {
-   link = "Comment",
-   default = true,
-})
+local ns = vim.api.nvim_create_namespace("SlangServerInactiveRegions")

 ---@param uri string
 ---@param ranges lsp.Range[]
@@ -21,7 +18,7 @@ local function apply_highlights(uri, ranges)
       vim.api.nvim_buf_set_extmark(bufnr, ns, range.start.line, range.start.character, {
          end_row = range["end"].line,
          end_col = range["end"].character,
-         hl_group = "SlangServerInactiveRegion",
+         hl_group = hl.INACTIVE_REGIONS,
          hl_eol = true,
       })
    end
 ~/slang-server   neovim-inactive
```

I get the error

```
Failed to run `config` for slang-server.nvim

...ver/clients/neovim/lua/slang-server/_core/highlights.lua:25: bad argument #1 to 'pairs' (table expected, got nil)

# stacktrace:
  - ~/slang-server/clients/neovim/lua/slang-server/_core/highlights.lua:25
  - ~/slang-server/clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua:3
  - ~/slang-server/clients/neovim/lua/slang-server/init.lua:4
  - /nix/store/6kppwggs0vn06xmisfw9abam65lrkyqq-neovim-unwrapped-0.12.2/share/nvim/runtime/filetype.lua:28
  - vim/_core/shared.lua:0
  - /nix/store/6kppwggs0vn06xmisfw9abam65lrkyqq-neovim-unwrapped-0.12.2/share/nvim/runtime/filetype.lua:27
  - vim/_core/editor.lua:0 _in_ **cmd**
  - /snacks.nvim/lua/snacks/picker/actions.lua:139 _in_ **jump**
  - /snacks.nvim/lua/snacks/explorer/actions.lua:326 _in_ **fn**
  - /snacks.nvim/lua/snacks/win.lua:362
```

clients/neovim/lua/slang-server/init.lua
evanwporter: > I think this races with the LSP initialization. Once the `slang-server` LSP client has started, updating its `capabilities` retroactively won't have any effect, and we can't guarantee when this plugin will be loaded relative to the client.
> 
> Possible solutions:
> 
> * Document that the user needs to add this to their own LSP config if they want to use this feature.
> * If the client is already running, update its capabilities and notify the server with `"workspace/didChangeConfiguration"` (if that's supported, @AndrewNolte?)
> 
> The former is obviously less than ideal. Thoughts, @toddstrader?

If it's really not possible to define the client capability in the plugin due to the load order, then I'd say option #1 is preferable (ie: `the user needs to add this to their own LSP config if they want to use this feature.`).

From my limited knowledge of LSP, it's not really possible to have the capabilities change after [initialization](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#initialize). 

While its certainty possible to update the [server side setting](https://github.com/hudson-trading/slang-server/blob/99484c1d5228b57dcb65a0ae65e311fba2b29c16/include/SlangLspClient.h#L17-L19) (and wouldn't be difficult if we make a custom command like `server/updateCapability`), it just seems to me to be a hacky method of implementing this.

evanwporter: I have to say though. I’m somewhat confused about how the plugin stuff/lsp load order works. I admit that it seems like it should be a problem especially with lazy loading, and yet I’ve had zero issue with this PR so far (using lazy.nvim).

I switched to neovim about a month ago so lot of this is still new to me.

Edit: Here's my config file for reference:

```
return {
  {
    "neovim/nvim-lspconfig",
    opts = {
      servers = {
        slang_server = {
          name = "slang-server",
          mason = false,
          cmd = { vim.fn.expand("~/slang-server/build/debug/bin/slang-server") },
          filetypes = { "systemverilog", "verilog", "slang" },
          single_file_support = true,
        },
      },
    },
  },
  {
    dir = vim.fn.expand("~/slang-server/clients/neovim/"),
    name = "slang-server.nvim",
    dependencies = {
      "MunifTanjim/nui.nvim",
    },
    ft = {
      "verilog",
      "systemverilog",
    },
    opts = {},
  },
}
```

clients/neovim/lua/slang-server/_lsp/experimental/inactive_regions.lua
sgaulter: That looks like the issue where `init.lua` hasn't been run and the config hasn't been initialised. Should be fixed by #399.

evanwporter:
@sgaulter

Hey Simon,

I did some research and I think the move is too follow in the footsteps of
rustaceanvim:

https://github.com/mrcjkb/rustaceanvim#zap-quick-setup

And have the neovim plugin completely own the setup for the lsp.

Most server side configuration is done with .slang/server.json anyways so I
don’t think the end user would be losing much here

It would (a) allow more flexibility in the future and (b) allow the plugin
to work out of the box.

The big issue is this is a breaking change and people would need to update
their configs (the pain could be mitigated by providing a warning if the
lsp client config exists).

Lmk what you think.

sgaulter:
Okay, I've done a bit more of a deep dive on this.

Tl;dr: lazy.nvim muddies the water, but we should be okay with the current ordering.

An LSP client is started on a FileType autocmd, which is created and registered by `vim.lsp.enable`.  The contents of `ftplugin/systemverilog.lua` will always get sourced before FileType autocmds run, so if you're not using lazy.nvim then it's always guaranteed the config gets loaded before the client is started ✅.

If the plugin is installed through lazy.nvim, **because it has no lazy triggers** (e.g., `ft=`, `cmd=` in its lazy spec), it gets added to the runtime path and `ftplugin/systemverilog.lua` sourced early when calling `lazy.setup` ✅.

Things would get murkier if a user decided to add a lazy trigger like `ft="systemverilog"` in their plugin spec, because in that case lazy.nvim creates a new FileType autocmd to add the plugin to the runtimepath on opening an SV file, and then the ordering in which that autocmd is registered relative to `vim.lsp.enable`'s isn't guaranteed ⚠️.

So to guarantee order, we should:

- Add `lazy=false` to the plugin spec in `lazy.lua` (which overrides any `ft` or other lazy triggers);
- Make the change suggested above to ensure the config gets sourced in `ftplugin/systemverilog.lua`.

sgaulter:
> And have the neovim plugin completely own the setup for the lsp.

I'm broadly opposed to going this direction, because a) it breaks convention from how users would expect to configure an LSP client (I do love banging on about "the principle of least surprise") and b) many users will have hooks like mason-lspconfig to auto-configure and enable servers that Mason installs, so we have to add a config burden on the user to opt slang-server out of that.

evanwporter:
See #352

PR #409, opened 2026-06-29
Add LSP commands for driver/load tracing

heyfey:
### New LSP commands
- `slang.getDriversWithLocation` / `slang.getLoadsWithLocation`:  return the RTL paths of the driver/load
  cone for a hierarchical path and its source location

### Refactor
- `ServerCompilation` no longer builds LSP CallHierarchy responses. The `getCallHierarchyCalls<P,R>` template is replaced by a direct `getConeLocations<isDrivers>()` returning `ConeEntry { path, location }`
- The incoming/outgoing call-hierarchy repackaging now lives in the LSP routes in `SlangServer.cpp`
- `getDrivers/LoadsWithLocation` return `ConeEntry` directly instead of synthesizing `CallHierarchyItem`


AndrewNolte:
Hey, are you able to get the precommit passing on this?

heyfey:
> Hey, are you able to get the precommit passing on this?

fixed clang-format, could you rerun workflow?

toddstrader:
Can you explain what this interface does that `prepareCallHierarchy` / `incomingCalls` / `outgoingCalls` does not?
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareCallHierarchy
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#callHierarchy_incomingCalls
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#callHierarchy_outgoingCalls

Using the preexisting LSP mechanism allows this to work without client specific code on the other end.  Also, do you really want to start from an RTL path?  I would imagine end users will want to start with `TextDocumentPositionParams`, AKA interacting with identifiers directly in the editor.  What is the end-to-end use case you're imagining that would start with knowing a fully qualified RTL path?

heyfey:
@toddstrader The purpose of the changes is to benefit other clients than slang-server, but using slang-server as backend. One of them is [sv-pathfinder](https://github.com/heyfey/sv-pathfinder), and the other I can think of might be an AI tool call to query driver/load, which are not editor-based. So yes, I do really want to start from an RTL path.

toddstrader:
I see.  Have you tried using `incomingCalls` or `outgoingCalls` as-is?  There's no state required from the `prepare` step so you can just skip it if it's not needed.

heyfey:
Yeah they're the first thing I tried but didn't work. Plus, I don't think it make sense to use standard LSP commands for non-IDE/editor applications.

src/SlangServer.cpp
```diff
@@ -112,6 +112,14 @@ lsp::InitializeResult SlangServer::getInitialize(const lsp::InitializeParams& pa
     registerCommand<waves::ItemToWaveform, std::monostate, &SlangServer::addToWaveform>(
         "slang.addToWaveform");
     registerCommand<std::string, std::monostate, &SlangServer::openWaveform>("slang.openWaveform");
+    registerCommand<std::string, std::vector<std::string>, &SlangServer::getDrivers>(
+        "slang.getDrivers");
+    registerCommand<std::string, std::vector<std::string>, &SlangServer::getLoads>(
+        "slang.getLoads");
```
AndrewNolte: Do we need to expose these too? Or are just the location ones needed

heyfey: Currently I only use the ones w locations. But I assume these are cleaner for clients that don't want locations

toddstrader: Yeah, suggest we minimize the surface area of the API which isn't locked down by the `slang-server` clients or the LSP itself.  It's great if you can get use out of the server without an LSP client, but that's specifically what `slang-server is and any extraneous API mechanisms are liable to be fragile.

heyfey: both dropped

AndrewNolte:
This branch is quite out of date. Do you mind giving me push access to this branch/PR? It looks like it's just main though, so there might be extra protection there

heyfey:
> This branch is quite out of date. Do you mind giving me push access to this branch/PR? It looks like it's just main though, so there might be extra protection there

Allow edits enabled, go ahead

AndrewNolte:
Accidentally pushed slang-server's main to the branch, then github closed the PR bc no commits, and so now I can't edit your branch 😭

AndrewNolte:
...

AndrewNolte:
So if you push I can reopen

AndrewNolte:
It's so odd it keeps giving me the option to 'reopen and comment', but then immediately closes it.

heyfey:
@AndrewNolte I can't reopen either. Creating #471

Issue #395, opened 2026-06-16
Client support: Kate and QT Creator


mdavidsaver:
Some feedback from a new user.  I have successfully built on a Debian 13 Linux host, and run with the [Kate](https://apps.kde.org/kate/) editor.  A promising start!

The [building](https://hudson-trading.github.io/slang-server/start/building/) document makes no mention of a Boost dependency.  Initially my build failed due to a missing header `boost/unordered/concurrent_flat_set.hpp`, which I found in the Debian package `libboost1.88-dev` (not in the default `libboost-dev`, which is boost 1.83).

```sh
sudo apt install libboost1.88-dev
```

With this additional package, my build succeeds.

`slang-server` seems to be working with KDE Kate 25.04.3.

Use with [Qt Creator](https://www.qt.io/development/tools/qt-creator-ide) fails with:

```
Failed to parse field 'capabilities': Failed to parse field 'textDocument': Failed to parse field 'codeAction': Failed to parse field 'codeActionLiteralSupport': Failed to parse field 'codeActionKind': Failed to parse field 'valueSet': Literal does not support string '*'. The following strings are supported: '', 'quickfix', 'refactor', 'refactor.extract', 'refactor.inline', 'refactor.rewrite', 'source', 'source.organizeImports', 'source.fixAll', 'notebook'.
```

The relevant section of the "initialize" which QtC 16.0.1 sends.

```json
{
    "jsonrpc": "2.0",
    "method": "initialize",
    "params": {
        "capabilities": {
            "textDocument": {
                "callHierarchy": {
                    "dynamicRegistration": true
                },
                "codeAction": {
                    "codeActionLiteralSupport": {
                        "codeActionKind": {
                            "valueSet": [
                                "*"
                            ]
                        }
                    }
                },
```

mdavidsaver:
["User Server Settings"](https://docs.kde.org/stable_kf6/en/kate/kate/kate-application-plugin-lspclient.html#lspclient-configuration) for Kate.

```json
{
    "servers": {
        "slang-server": {
            "command": [ "/home/dev/source/slang-server/build/bin/slang-server"],
            "highlightingModeRegex": "Verilog",
            "documentLanguageId": false,
            "rootIndicationFileNames": [".slang"]
        }
    }
}
```

Use `rootIndicationFileNames` as I have no `.kateproject` file.

[...]

PR #410, opened 2026-07-06
Fixes for Kate and QtCreator editors

mdavidsaver:
Address issues from #395.

- Accept `CodeActionKind` `"*"`, as delivered by QtCreator
- Accept `LanguageKind` with empty string, as delivered by Kate
- Respect `textDocument.definition.linkSupport`.  Both Kate and QtC omit `linkSupport`, and in turn expect `Location` to be returned.

For the last.  I took `capabilities.experimental` as a template.  Since `Location` is practically a sub-set of `LocationLink`, I added a back translation in `SlangServer::getDocDefinition()`.  I also changed (I think) the test client to pass `linkSupport=true` to avoid breaking any existing tests.

### DIFF
--- a/src/SlangServer.cpp
+++ b/src/SlangServer.cpp
@@ -164,6 +164,12 @@ lsp::InitializeResult SlangServer::getInitialize(const lsp::InitializeParams& pa
 
     loadConfig();
 
+    if (const auto& tDoc = params.capabilities.textDocument) {
+        if (const auto& tDef = tDoc->definition) {
+            m_client.linkSupport = tDef->linkSupport.value_or(false);
+        }
+    }
+
     if (params.capabilities.experimental) {
         auto exp = rfl::from_generic<lsp::ExperimentalClientCapabilities>(
             *params.capabilities.experimental);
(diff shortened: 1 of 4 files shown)


src/SlangServer.cpp
```diff
@@ -164,6 +164,12 @@ lsp::InitializeResult SlangServer::getInitialize(const lsp::InitializeParams& pa
 
     loadConfig();
 
+    if (const auto& tDoc = params.capabilities.textDocument) {
+        if (const auto& tDef = tDoc->definition) {
+            m_client.linkSupport = tDef->linkSupport.value_or(false);
```
AndrewNolte, line 169: Let's just copy the whole capabilities object

tests/system/conftest.py
```diff
@@ -268,7 +268,13 @@ async def client(my_options: Flags) -> AsyncGenerator[SlangClient, None]:
 
     response = await client.initialize_async(
         types.InitializeParams(
-            capabilities=types.ClientCapabilities(),  # Ignored for now
+            capabilities=types.ClientCapabilities(
+                text_document=types.TextDocumentClientCapabilities(
+                    definition=types.DefinitionClientCapabilities(
+                        link_support=True,
```
AndrewNolte, line 274: Can we also get a new test showing that we have the different object being returned for this route? with link support off?

src/SlangServer.cpp
mdavidsaver: Done, I think.  The result seems less efficient, with `capabilities` being inspected on every call.

tests/system/conftest.py
mdavidsaver: I have not yet had time to investigate how/where to do this.  Can you suggest a starting point?

AndrewNolte:
Merged https://github.com/hudson-trading/slang-server/pull/481 from these changes

PR #432, opened 2026-07-26
alpine builds

### LINKED ISSUE #418: musl releases
Can you add musl releases? If this isn't done by this weekend, then i'll open a pr

evanwporter:
closes #418


AndrewNolte:
I think this can be added without having `shell` added to each matrix entry - it's just the one that has uses a non default shell.sh

PR #433, opened 2026-07-26
Revert "cancel-in-progress on pull request pushes"

evanwporter:
Reverts hudson-trading/slang-server#386

This is kind of annoying to have builds cancel midway through because you pushed a new commit or not be able to run a new workflow since an old one is running.

I don't really see the point since github foots the workflow bill.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,10 +32,6 @@ on:
       - 'pyproject.toml'
       - 'uv.lock'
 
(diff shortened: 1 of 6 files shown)


AndrewNolte:
There was further discussion on https://github.com/hudson-trading/slang-server/pull/386, closing for now. I'd rather the newer ci just finish faster than *maybe* finding errors faster if the old ci finishes.

PR #435, opened 2026-08-05
Fix macro hover showing following typedef instead of `define

### LINKED ISSUE #425: Macro hover renders the typedef following the `define instead of the macro definition
**Version:** v0.2.8 (also reproduces on main @ 4f33c99)

## Summary

Hovering a macro usage shows an unrelated `typedef` from the defining file in
the "definition" code block — specifically, whatever typedef comes *after* the
`` `define `` in the source. In typedef-heavy header files (very common: width
macros above struct definitions), every macro hover shows a big struct instead
of its own definition.

## Reproduction

`defs.svh`:

```systemverilog
`define DATA_W 32

typedef struct packed {
  logic [7:0]  field_a;
  logic [15:0] field_b;
} big_struct_t;

`define A 8
`define B 4
typedef struct packed { logic x; } small_t;
```

Hover any usage of `` `DATA_W `` → the code block shows `big_struct_t`, not
`` `define DATA_W 32 ``. Hover `` `A `` or `` `B `` → both show `small_t`.
The "Expands to" section is correct; only the definition block is wrong.

## Root cause

`selectDisplayNode()` in `src/util/Formatting.cpp` ends with:

```cpp
if (fmtNode->parent && fmtNode->parent->kind == syntax::SyntaxKind::TypedefDeclaration) {
    fmtNode = fmtNode->parent;
}
```

This promotion is meant for nodes like `EnumType` so the full typedef is
displayed. But directives are trivia attached to the next real token, so a
`DefineDirectiveSyntax`'s `parent` is the declaration that *follows* it in the
file. When that declaration is a typedef, the display node for
`MacroTarget::getHover()` (via `SyntaxTarget::renderCode`) silently becomes
that typedef, and the hover renders it — including its harvested doc comments —
in place of the `` `define ``.

## Suggested fix

Skip the typedef-parent promotion (and arguably the whole `selectDisplayNode`
adjustment) when `node.kind == SyntaxKind::DefineDirective`, e.g. early-return
the node itself for directive kinds.

adxb-pxthxn:
## Summary
- Fixes #425: hovering a macro usage no longer renders a following `typedef` in the definition code block.
- `selectDisplayNode()` early-returns for `DefineDirective` so trivia-parent promotion to `TypedefDeclaration` does not apply.
- Adds `HoverMacroNotFollowingTypedef` regression coverage.
## Test plan
- [x] `./build/bin/server_unittests 'HoverMacro*'` (passes)
- [x] Manually hover `` `DATA_W `` / `` `A `` / `` `B `` in a file with trailing typedefs and confirm the hover code block shows the `` `define ``, while "Expands to" stays correct

### DIFF
--- a/src/util/Formatting.cpp
+++ b/src/util/Formatting.cpp
@@ -352,6 +352,12 @@ lsp::MarkupContent svCodeBlock(const std::string_view code) {
 }
 
 const syntax::SyntaxNode& selectDisplayNode(const syntax::SyntaxNode& node) {
+    // Directives live as trivia on the following token, so their syntax parent is the
+    // next declaration in the file — not a semantic parent. Promoting to a following
+    // TypedefDeclaration would show the wrong node on macro hover.
+    if (node.kind == syntax::SyntaxKind::DefineDirective)
+        return node;
+
     const syntax::SyntaxNode* fmtNode = &node;
     switch (node.kind) {
         // Adjust these to just be the header
--- a/src/util/Formatting.cpp
+++ b/src/util/Formatting.cpp
@@ -352,14 +352,12 @@ lsp::MarkupContent svCodeBlock(const std::string_view code) {
 }
 
 const syntax::SyntaxNode& selectDisplayNode(const syntax::SyntaxNode& node) {
-    // Directives live as trivia on the following token, so their syntax parent is the
-    // next declaration in the file — not a semantic parent. Promoting to a following
-    // TypedefDeclaration would show the wrong node on macro hover.
-    if (node.kind == syntax::SyntaxKind::DefineDirective)
-        return node;
-
     const syntax::SyntaxNode* fmtNode = &node;
     switch (node.kind) {
+        // Directives live as trivia on the following token, so their syntax parent is the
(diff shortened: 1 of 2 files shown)


adxb-pxthxn:
Screenshot_2026-08-05_at_3 44 45_pm-b89cfd6a-e5b8-
Screenshot_2026-08-05_at_3 44 54_pm-4daf9b8a-6f3b-

src/util/Formatting.cpp
```diff
@@ -352,6 +352,12 @@ lsp::MarkupContent svCodeBlock(const std::string_view code) {
 }
 
 const syntax::SyntaxNode& selectDisplayNode(const syntax::SyntaxNode& node) {
+    // Directives live as trivia on the following token, so their syntax parent is the
+    // next declaration in the file — not a semantic parent. Promoting to a following
+    // TypedefDeclaration would show the wrong node on macro hover.
+    if (node.kind == syntax::SyntaxKind::DefineDirective)
+        return node;
```
AndrewNolte: This can be part of the switch stmt

adxb-pxthxn: Sweet, I have made the changes, tested and pushed.

PR #436, opened 2026-08-06
Implement initial semantic token enchancement

adxb-pxthxn:
## Summary
- add Phase 1 LSP semantic tokens: classify slang tokens using `TokenKind` + declaration/`SyntaxIndexer` context (no eager full-document symbol resolution on edit).
- Advertise `textDocument/semanticTokens/full` and return delta-encoded tokens from `ShallowAnalysis`.
- Golden-test the high-value cases TextMate gets wrong: parameters, typedef names, system tasks.
- Parent-syntax fallback uses an explicit `SyntaxKind` switch (not string matching on kind names). Declaration containers like `ModuleHeader` / `ClassDeclaration` only classify when the token is that declaration’s name.
## Motivation
Syntax highlighting today is client TextMate. That fails for user types (`word_t`), parameters (`WIDTH`, `ENTRY_T`), and similar cases (#268). Semantic tokens let the server, which already has slang’s tokenizer and shallow analysis, drive accurate highlighting.
## Approach (overall roadmap)
**Phase 1 (this PR)**: declaration-aware classification
- Walk existing `SyntaxIndexer::collected` (request path only).
- Lexical map: keyword / number / string / operator / macro.
- Identifiers: `SymbolIndexer` for declarations in-buffer, explicit `SyntaxKind` parent walk as fallback.
- Keep TextMate. Semantic tokens augment (client theme still owns colors).
**Phase 2**: Real semantics on references
- `getSymbolAtToken` for uses (`pkg::thing`, typedef uses, etc.).
- Richer modifiers, optional range/delta, request cancellation if analysis is mid-rebuild.
## Efficiency notes
- Does **not** classify on `onChange`. Analysis is still invalidated as today and rebuilt lazily.
- Token stream is a single linear pass over already-collected tokens + O(1) declaration map lookups.
- Syntax fallback is an enum switch over kinds, not string search on kind names.
## Test plan
- [x] `build/bin/server_unittests SemanticTokens_Phase1`
- [ ] Spot-check in editor: typedef / parameter names pick up semantic colors while TextMate still handles punctuation
- [ ] Confirm typing latency unchanged (tokens only computed on semantic-token requests)
Partially addresses #268

### DIFF
--- a/src/document/ShallowAnalysis.cpp
+++ b/src/document/ShallowAnalysis.cpp
@@ -13,7 +13,6 @@
 #include "util/Converters.h"
 #include "util/Logging.h"
 #include "util/SlangExtensions.h"
-#include <array>
 #include <fmt/format.h>
 #include <memory>
 #include <string_view>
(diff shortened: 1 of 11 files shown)


src/document/ShallowAnalysis.cpp
```diff
@@ -56,6 +58,406 @@ static bool symbolsMatch(const ast::Symbol* a, const ast::Symbol* b) {
     }
     return false;
 }
+
+namespace {
+enum class SemanticTokenTypeIndex : lsp::uint {
```
AndrewNolte: This should be able to fit into uint8; let's also try to inline the "none" option here. There's also some rfl thing for mapping enum to string directly at serialization time. I'm concerned about merging semantic tokens bc it's a heavy lsp method with I would argue not much benefit. I'd like to merge threaded/cancellable requests and some latency measurements before doing semantic tokens. If it's behind a config option that's off by default I think that'd be fine too.

adxb-pxthxn: Thanks, addressed all three: the enum is `uint8_t` now (it's the underlying type of the spec's `rfl::Literal`, so it follows from that rather than being hand-set), the `none` case is an inline enum sentinel instead of `std::optional`, and the legend is derived from `SemanticTokenTypesOptions` via `value_of<>`/`strings()`, so the two hand-maintained string arrays are gone and a bad token name is now a compile error.

That last change means we advertise the full 24-type spec legend rather than our own 13, and the wire indices are now the spec's. Names are unchanged, so nothing client-visible shifts.

Also put it behind a config option that's **off by default**, so it's inert unless someone opts in. I'll pick up threaded/cancellable requests and latency measurements next.

PR #444, opened 2026-08-07
Add request latency instrumentation and microbench

adxb-pxthxn:
## Summary
- Instrument JSON-RPC request/notification handling with elapsed-ms timing; capture under the RPC mutex, format/log only after unlock so stderr I/O is not on the critical path.
- Centralize `elapsedMs` in `util/Timing.h` (no logging macros) and reuse it from RPC logging and `ScopedTimer`.
- Add a Catch `[latency]` microbench that times true first-touch via `didOpen` (`open_and_analyze`), analysis rebuild after edit, token walk, documentSymbol, hover, and inlayHint.

Semantic tokens (`#436` / `#268`) add a whole-document request path. Review feedback on that work was clear: do not enable it by default until we have **measured latency** and a plan for **threaded/cancellable** requests. Guessing whether `semanticTokens/full` (or any other handler) is “cheap enough” is the wrong bar for this codebase.

This PR is the measurement step of that sequence:

1. **Dogfood in the editor** — every request/notification finish line now includes elapsed ms, so real sessions show which handlers dominate typing/edit latency.
2. **Reproducible baseline in CI** — the microbench locks in numbers for open/analyze, rebuild-after-edit, and common whole-doc paths, so later threading or semantic-token work can be judged against evidence rather than feel.
3. **Decide what to build next** — if whole-doc work stays sub-ms–few-ms on representative sizes, cooperative cancellation may be enough; if open/rebuild or classify is heavy, we have justification for offloading. Threading without these numbers would be premature architecture.

The instrumentation is deliberately low-perturbation: under the RPC mutex we only run the handler, take clock samples, and send the response (stdout still needs the lock for WCP). Formatting to stderr happens **after** unlock so logging I/O does not stretch the critical section.

## Test plan
- [ ] `cmake --build build -j8 --target server_unittests`
- [ ] `build/bin/server_unittests "RequestLatency_LargeDocument"` — look for `LATENCY` lines (`{:.3f}ms`)
- [ ] Point an editor at `build/bin/slang-server` and confirm finish lines like `---> textDocument/hover <id> (0.047ms)`
- [ ] Confirm normal request handling is unchanged when ignoring the new timing logs

## Notes
- First-touch is measured around `openFile` (diags force `getAnalysis`), not a later warm cache hit.
- Soft upper bounds in the microbench are catastrophic-regression guards only, not performance SLOs.
- Follow-ups: semantic-token classify timing once `#436` lands; then cancellation / threading only if the numbers say so.

### DIFF
--- a/include/lsp/JsonRpcServer.h
+++ b/include/lsp/JsonRpcServer.h
@@ -12,7 +12,6 @@
 #include "lsp/LspTypes.h"
 #include "rfl/Generic.hpp"
 #include "util/Timing.h"
-#include <chrono>
 #include <concepts>
 #include <cstdint>
 #include <fmt/format.h>
@@ -160,7 +159,7 @@ protected:
             // Notification
             auto it = notifications.find(request.method);
             if (it != notifications.end()) {
-                const auto start = std::chrono::steady_clock::now();
+                ScopedElapsedMs elapsed(timing.ms);
                 try {
                     if (request.params.has_value()) {
                         it->second(request.params.value());
@@ -168,11 +167,9 @@ protected:
                     else {
                         it->second(std::nullopt);
                     }
-                    timing.ms = elapsedMs(start);
                     timing.kind = HandlerTiming::Kind::NotificationOk;
                 }
                 catch (const std::exception& e) {
-                    timing.ms = elapsedMs(start);
                     timing.error = e.what();
                     timing.kind = HandlerTiming::Kind::NotificationError;
                 }
@@ -205,7 +202,7 @@ protected:
         auto it = requests.find(request.method);
 
         if (it != requests.end()) {
-            const auto start = std::chrono::steady_clock::now();
+            ScopedElapsedMs elapsed(timing.ms);
             try {
                 rfl::Generic req_response;
                 if (request.params.has_value()) {
@@ -214,12 +211,10 @@ protected:
                 else {
                     req_response = it->second(rfl::Generic{});
                 }
-                timing.ms = elapsedMs(start);
                 timing.kind = HandlerTiming::Kind::RequestOk;
                 return req_response;
             }
             catch (const std::exception& e) {
-                timing.ms = elapsedMs(start);
                 timing.error = e.what();
                 timing.kind = HandlerTiming::Kind::RequestError;
                 return RpcError{.code = 1, .message = e.what()};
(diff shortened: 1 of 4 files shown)


include/lsp/JsonRpcServer.h
```diff
@@ -24,6 +27,57 @@
...
+    };
+
+    Kind kind = Kind::None;
+    std::string method;
+    std::string id;
+    std::string error;
+    double ms = 0;
+};
+
+inline void logHandlerTiming(const HandlerTiming& timing) {
```
AndrewNolte, line 55: lots of duplicate code here wth the timing.ms and the \n\n

include/lsp/JsonRpcServer.h
```diff
@@ -24,6 +27,57 @@
 
 namespace lsp {
 
+namespace {
+/// Captured under the RPC mutex; formatted only after unlock so logging I/O is
+/// not serialized onto the critical path.
+struct HandlerTiming {
```
AndrewNolte, line 33: why not also use raii to get the latency?

include/lsp/JsonRpcServer.h
```diff
@@ -24,6 +27,57 @@
...
+        NotificationOk,
+        NotificationError,
+        RequestOk,
+        RequestError,
+        Ignored,
+        NotFound,
+    };
+
+    Kind kind = Kind::None;
+    std::string method;
```
AndrewNolte: this and others can be string_view

include/lsp/JsonRpcServer.h
adxb-pxthxn: Yep, I have cleaned that up.

include/lsp/JsonRpcServer.h
adxb-pxthxn: Yep, done in 10ab6e0.

Added a silent ScopedElapsedMs in Timing.h that writes into timing.ms on destruction. Didn’t reuse ScopedTimer here because that logs on destroy, and we still want stderr formatting deferred until after the RPC mutex unlock.

include/lsp/JsonRpcServer.h
adxb-pxthxn: Done, method and string ids are now string_views into the request (which outlives logging in handleMessage).

AndrewNolte:
merged https://github.com/hudson-trading/slang-server/pull/475

PR #445, opened 2026-08-09
Fix LSP positions after `line directives

### LINKED ISSUE #424: documentSymbol positions honor `line directives, producing out-of-range LSP positions
**Version:** slang-server 0.2.8+ (also present on current `main`, 4f33c99)

## Summary

LSP positions are computed with `SourceManager::getLineNumber()`, which honors `` `line `` directives — so for files containing `` `line `` (common in generated/preprocessed RTL, e.g. register-block generators), `textDocument/documentSymbol` returns positions at the *directive-remapped* line number instead of the physical line in the file. Clients then jump/set-cursor out of the buffer's range and crash (e.g. trouble.nvim: `Invalid cursor line: out of range`).

Per the LSP spec, `Position` is a zero-based offset into the document *text* — it must be the physical line, regardless of any `` `line `` remapping (which is only meant for diagnostics-style reporting).

## Repro

```systemverilog
module linedir(input logic clk);
`line 500 "linedir.sv" 0
  logic delta;
endmodule
```

`textDocument/documentSymbol` response for this 4-line file:

```
linedir  sel=[0,7]
  clk    sel=[0,27]
  delta  sel=[499,8]    <-- line 499 in a 4-line document
```

## Cause

`toPosition()` in `src/util/Converters.cpp` uses `sourceManager.getLineNumber(loc)`, which applies `LineDirectiveInfo` adjustments. Every LSP position produced through `toPosition`/`toRange` is affected, so the same physical-vs-reported mismatch presumably also applies to definitions/references/diagnostics that land inside a `` `line ``-remapped region.

Note slang's `getRawLineNumber` is private, so the fix likely needs either a small upstream slang change (expose a physical-line query) or computing the line from the buffer contents/offsets directly in the server.

Nehal70:
## Summary
- `toPosition` / `toRange` now emit **physical** buffer line numbers instead of `` `line ``-remapped lines from `SourceManager::getLineNumber`, so LSP ranges stay in-document.
- Wildcard inlay-hint same-line checks use the same physical mapping.
- Catch2 regressions cover documentSymbol, goto definition, diagnostics, and documentHighlight for the linedir repro (`delta` at physical line 2, not 499).
- Test harness `getLspLocation` now uses the same physical conversion as production.

## Issue
Fixes #424

## Test plan
- [x] Added/updated tests: `tests/cpp/DocumentSymbolTests.cpp`, `tests/cpp/utils/ServerHarness.cpp`
- [x] Reproduced failure on tip of `main` before fix (`selectionRange.start.line == 499`)
- [x] Ran:
  ```bash
  git submodule update --init --recursive
  cmake -B build -DCMAKE_BUILD_TYPE=Debug -DCMAKE_CXX_COMPILER=clang++
  cmake --build build -j4 --target server_unittests
  ./build/bin/server_unittests '*LineDirective*'
  ./build/bin/server_unittests
  ctest --test-dir build --output-on-failure
  ```
- [x] Result: line-directive suite `All tests passed (41 assertions in 4 test cases)`; full `server_unittests` `All tests passed (58825 assertions in 210 test cases)`

## Notes
- OS: Ubuntu 24.04 WSL2
- Physical lines use the same newline rules as slang's `getRawLineNumber` (macros expanded; `` `line `` ignored). Exposing `getRawLineNumber` publicly in the slang fork would be a nice follow-up to reuse cached offsets.
- CI workflows are currently `action_required` for this first-time contributor fork PR; please approve workflow runs when convenient.

### DIFF
--- a/src/document/InlayHintCollector.cpp
+++ b/src/document/InlayHintCollector.cpp
@@ -188,7 +188,8 @@ void InlayHintCollector::handle(const HierarchyInstantiationSyntax& syntax) {
                                     FORMATTING_INDENT - 1;
 
                     auto getLine = [&](const auto& loc) {
-                        return m_analysis.m_sourceManager.getLineNumber(loc);
+                        // Compare physical document lines (ignore `line remapping).
+                        return toPosition(loc, m_analysis.m_sourceManager).line;
                     };
 
                     if (getLine(instanceSyntax->openParen.location()) ==
--- a/src/util/Converters.cpp
+++ b/src/util/Converters.cpp
@@ -7,9 +7,7 @@
 //------------------------------------------------------------------------------
 #include "util/Converters.h"
 
-#include <algorithm>
 #include <fmt/format.h>
-#include <vector>
 
 #include "slang/text/SourceLocation.h"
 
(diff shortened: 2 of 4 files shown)


Nehal70:
Added a follow-up commit with broader edge-case coverage for `` `line `` physical positions:

- no-`` `line `` control
- multiple `` `line `` directives
- CRLF buffers
- `toPosition` ↔ `toSourceLocation` round-trip
- references + hover
- macro expansion after `` `line ``

Local: `server_unittests '[line_directive]'` → 11 cases / 114 assertions passed; full suite also green.

AndrewNolte:
This requires exposing some more data from the source manager- we don't want to do a bunch of post-processing outside the sm when the data is there on a shorter path.

I'll patch the slang fork to expose the raw line number and fix this. It would be nice as a followup to also publish diagnostics for the original source file in addition to the generated source file, and expose hovers/gotos for line directives. I'll also patch to expose the directive as queryable per file for that purpose.

Nehal70:
> This requires exposing some more data from the source manager- we don't want to do a bunch of post-processing outside the sm when the data is there on a shorter path.
> 
> I'll patch the slang fork to expose the raw line number and fix this. It would be nice as a followup to also publish diagnostics for the original source file in addition to the generated source file, and expose hovers/gotos for line directives. I'll also patch to expose the directive as queryable per file for that purpose.

Thanks for the feedback. That makes sense and is the right path forward. Happy to take a crack at that approach if useful. Are you already mid-flight on this, or would you like me to open that PR? Would also be interested in working on the diagnostics follow up you mentioned once this is landed (either by you or me).

PR #446, opened 2026-08-09
CI: try to install lua not from lua.org

AndrewNolte:
Stacked PRs:
 * __->__#446
 * #442

--- --- ---


AndrewNolte:
Abandoning for now; probably best to modify the neovim action instead

PR #448, opened 2026-08-12
Assign neovim plugin a version

evanwporter:
In my own neovim config I'd prefer installing plugins with my package manager (nix).

https://github.com/evanwporter/dotfiles/blob/9e9f3cd8a2f21b49e94038ef0244be988f16c496/nix/home/modules/neovim.nix#L87-L118

Unfortunately because the slang-server.nvim plugin has no versions its not really suitable to be packaged.

Thus this PR assigns the very first version to the neovim plugin (0.2.1) and also it changes the ci script to detect version changes in the rockspec and assign a corresponding tag to the repo:

https://github.com/hudson-trading/slang-server.nvim

The hope is that like the vscode extension, the slang-server plugin can be updated independently of the language server.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
         env:
           SLANG_SERVER_BIN: ../../build/${{ matrix.preset }}/bin/slang-server
         with:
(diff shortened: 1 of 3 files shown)


evanwporter:
No longer needed; I realized I could push a plugin without a version

https://github.com/NixOS/nixpkgs/pull/553368

PR #449, opened 2026-08-12
[Neovim] Handle `textdocument/inactiveRegions`

evanwporter:
@sgaulter many moons have passed since #402 and I now feel much more comfortable configuring neovim.

I understand now the concerns you raised and I also understand the lua that chatgpt outputs XD.

In this version, inactive-region support is strictly opt-in, as described in the documentation I added. Users must explicitly advertise the capability and register the corresponding handler.

If they do not opt in, there is no additional runtime overhead: the inactive-region implementation is loaded only when configured.

This may not be the most conventional approach, but unless a user explicitly opts in, the feature is effectively invisible.

image

Edit: the reason why I'm going with this approach is because semantic tokens requires way more work.

Edit 2: I realized that this config could just occupy a section of the documentation instead.


sgaulter:
I remember on the previous version you mentioned there were errors when rapidly deleting lines of code. Does this still suffer from that?

I ask because I assumed the problem was race conditions between the highlights being applied and the source changing, which I would think might still be the case here.

evanwporter:
> I remember on the previous version you mentioned there were errors when rapidly deleting lines of code. Does this still suffer from that?
>
> I ask because I assumed the problem was race conditions between the highlights being applied and the source changing, which I would think might still be the case here.

Sorry for the delay. To be honest I haven't encountered that issue since. But I also haven't changed very much from the original pr so its probably still an issue here.

This is probably a good reason  to not merge this pr.

My big realization here is that I can access the feature I want (inactive regions) without having to maintain a fork, so I'm happy.

PR #451, opened 2026-08-12
add support for `textdocument/typeDefinition`

evanwporter:
Instead of `go to definition`, then `go to definition` on the type, we can now directly go to the type.

### DIFF
--- a/tests/cpp/GotoTests.cpp
+++ b/tests/cpp/GotoTests.cpp
@@ -1,6 +1,7 @@
 // SPDX-FileCopyrightText: Hudson River Trading
 // SPDX-License-Identifier: MIT
 
+#include "utils/GoldenTest.h"
 #include "utils/ServerHarness.h"
 
 using namespace slang;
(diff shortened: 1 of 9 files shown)


tests/cpp/GotoTests.cpp
```diff
@@ -111,6 +123,44 @@ TEST_CASE("GotoDefinition_IfdefUndefinedMacro") {
...
+    // Invoke go-to-type-definition on the variable names. Both should select the
+    // underlying typedef name, including through the unpacked array wrapper.
+    auto recordTypeDefinition = [&](std::string label, std::vector<lsp::LocationLink> links) {
+        REQUIRE(links.size() == 1);
+        const auto& start = links[0].targetSelectionRange.start;
+        auto line = doc.getLine(start.line + 1);
+        if (line.ends_with('\n'))
+            line.remove_suffix(1);
+        golden.record(std::move(label),
+                      TypeDefinitionGoldenResult{.destinationLine = std::string(line),
```
AndrewNolte: Can we render the LocationLink instead of just marking the line/char/file? We can also make this cover more cases if we run/render per sym on larger file, like how the hover golden outputs work.

evanwporter: lmk if this doesn't address your second concern.

Edit: i re read what you said and it does not

evanwporter:
What did you change? I can’t tell because you squashed everything

AndrewNolte:
Just rebased

AndrewNolte:
They have a 'compare' button now

Screenshot 2026-08-18 at 10 01 59 AM

AndrewNolte:
Some gaps I noticed:
- instances -> module def
- typdef -> should be same as goto (in the case of some_t[$], it goes to some_t which seems wrong)
- enum -> enum def, so also same as goto

evanwporter:
okay ill try and fix it this weekend.

evanwporter:
I don't feel good about merging without more testing because most of this is AI generated. But it is ready for review I think

AndrewNolte:
Lets put the pure utility functions in SlangExtensions and it should be good to merge - it's not like it can harm other routes. 

I also want to move away from the raw json dumping for snapshot test outputs- especially if it's dumping a source loc. Snapshot tests should ideally be easily verifiable, and dumping json with line/col numbers is not. I'm thinking about converting those to rendered outputs except for maybe the hierarchy tests.

evanwporter:
These JSON tests have the snapshot / outputs.

I agree though about the lines columns not being easily verifiable—but ideally you only need to verify once.

Are you thinking the snapshots should be like the hover tests where it’s just a text with output

PR #452, opened 2026-08-13
Fix call hierarchy item ranges and duplicates

### LINKED ISSUE #423: Call hierarchy items have (0,0) range/selectionRange and undeduplicated cone leaves
**Version:** slang-server 0.2.8+ (also present on current `main`, 4f33c99)
**Client:** Neovim 0.12 + trouble.nvim (but the problem is client-independent — see below)

## Summary

`callHierarchy/incomingCalls` / `outgoingCalls` (driver/load cone tracing) return `CallHierarchyItem`s whose `range` and `selectionRange` are left default-constructed at `(0,0)-(0,0)`, and the response contains one entry per cone **leaf** with no deduplication — so a multi-bit signal repeats the same driver location several times.

Both come from `getCallHierarchyCalls` in `include/ast/ServerCompilation.h` (line 74): the pushed items fill only `.name` and `.uri` (lines 88/92), and unlike its sibling `getConePaths` (which dedups via a `seen` set, line 106), the loop has no dedup.

## Repro

```systemverilog
///////////////////////////////////////////////////////////////////////
// banner comment — line 1 on purpose
///////////////////////////////////////////////////////////////////////
module top(
  input  logic [3:0] a,
  input  logic [3:0] b,
  input  logic       clk,
  output logic [3:0] y
);
  logic [3:0] x;
  assign x = a & b;
  always_ff @(posedge clk) y <= x;
endmodule
```

With a build (`--top top`) configured, request `textDocument/prepareCallHierarchy` on `x` (line 11) followed by `callHierarchy/incomingCalls`. The raw response items:

```
name=top.a  fromRanges=[[10,13]]  from.range=[0,0]-[0,0]  from.selectionRange=[0,0]-[0,0]
name=top.b  fromRanges=[[10,17]]  from.range=[0,0]-[0,0]  from.selectionRange=[0,0]-[0,0]
```

`fromRanges` is correct; `from.range`/`from.selectionRange` are not. Per the LSP spec both fields are **required** on `CallHierarchyItem`, and `selectionRange` is what clients use to represent/jump to the caller item itself.

## Why this matters in practice

Clients that render the hierarchy label each entry using the *item's* range. trouble.nvim, for example, reads the text at `from.range` for the entry label — i.e. **line 1 of the driving file**. Since RTL files commonly start with a `/////…` banner, every driver in the panel renders as a wall of slashes:

```
/////////////////////////////////////////////////////////////////////// [753, 30] slang-server
/////////////////////////////////////////////////////////////////////// [753, 30] slang-server
/////////////////////////////////////////////////////////////////////// [753, 30] slang-server
/////////////////////////////////////////////////////////////////////// [753, 47] slang-server
...
```

…with each location repeated once per cone leaf (multi-bit drivers), even though every repeat carries identical `name` + `fromRanges`.

## Suggested fix

In `getCallHierarchyCalls`:

1. Set `range`/`selectionRange` on the `from`/`to` item to the leaf's source range (the same range already emitted in `fromRanges`), so clients have a valid location for the item itself.
2. Dedup entries the same way `getConePaths` does (e.g. by `hier` + range), or aggregate repeated leaves into one entry with multiple `fromRanges`.

Happy to send a PR if useful.

koriyoshi2041:
Closes #423.

Call hierarchy results now populate each item's required `range` and `selectionRange` from the cone leaf source range. Duplicate cone leaves are collapsed by hierarchical path, file, and source range, while distinct references to the same signal remain separate.

The shared call hierarchy test helper now checks item ranges and exact result cardinality for both incoming and outgoing calls.

Tests:
- `./build/bin/server_unittests "Cone Tracing"`
- `./build/bin/server_unittests` (211 cases, 58,886 assertions)
- clang-format on changed files
- `git diff --check`

### DIFF
--- a/clients/neovim/spec/plugin_spec.lua
+++ b/clients/neovim/spec/plugin_spec.lua
@@ -162,6 +162,7 @@ describe("SlangServer", function()
       end
       assert(item)
       assert.are.same("foo.a", item.name)
+      assert.are_not.same(item.selectionRange.start, item.selectionRange["end"])
 
       local outgoing = vim.lsp.buf_request_sync(0, "callHierarchy/outgoingCalls", { item = item }, 5000)
       assert(outgoing)
--- a/clients/neovim/spec/plugin_spec.lua
+++ b/clients/neovim/spec/plugin_spec.lua
@@ -143,6 +143,42 @@ describe("SlangServer", function()
       assert.are.same(expected, table.concat(lines, "\n"))
       vim.api.nvim_buf_delete(0, { force = true })
    end)
+
+   it("Vanilla call hierarchy keeps outgoing loads navigable", function()
+      local line = vim.api.nvim_buf_get_lines(0, 3, 4, false)[1]
+      local character = assert(string.find(line, "a;", 1, true)) - 1
+      local prepare = vim.lsp.buf_request_sync(0, "textDocument/prepareCallHierarchy", {
+         textDocument = { uri = vim.uri_from_bufnr(0) },
+         position = { line = 3, character = character },
+      }, 5000)
+      assert(prepare)
+
+      local item
+      for _, response in pairs(prepare) do
+         if response.result and response.result[1] then
+            item = response.result[1]
+            break
+         end
+      end
+      assert(item)
+      assert.are.same("foo.a", item.name)
+
+      local outgoing = vim.lsp.buf_request_sync(0, "callHierarchy/outgoingCalls", { item = item }, 5000)
+      assert(outgoing)
+
+      local calls
+      for _, response in pairs(outgoing) do
+         if response.result then
+            calls = response.result
+            break
+         end
+      end
+      assert(calls)
+      assert.are.same(1, #calls)
+      assert.are.same("foo.b", calls[1].to.name)
+      assert.are.same(item.uri, calls[1].to.uri)
+      assert.are.same(item.selectionRange, calls[1].fromRanges[1])
+   end)
 end)
 
 -- TODO (tests)
--- a/clients/neovim/tests/foo.sv
+++ b/clients/neovim/tests/foo.sv
@@ -1,4 +1,8 @@
 module foo;
+    logic a;
+    logic b;
+    assign b = a;
+
     for(genvar i = 0; i < 4; i++) begin: gen_loop
         sub #(.param(i)) the_sub();
     end
--- a/include/ast/ServerCompilation.h
+++ b/include/ast/ServerCompilation.h
@@ -13,6 +13,8 @@
 #include "util/Converters.h"
 #include <filesystem>
 #include <memory>
+#include <set>
+#include <tuple>
 #include <vector>
 
 #include "slang/util/Bag.h"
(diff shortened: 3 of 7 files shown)


toddstrader:
@koriyoshi2041 / @somyadashora can you be more explicit about what's happening with trouble.nvim?  I've only ever tested this with the stock neovim interface:
https://neovim.io/doc/user/lsp/#vim.lsp.buf.incoming_calls()

Could you show some complete example JSON?  Also I don't quite understand what's going on with the "wall of slashes" example.  I gather trouble.nvim is showing a preview of the drivers.  But I don't grok where those numbers are coming from.

@koriyoshi2041 can you add a test which would fail without this to demonstrate the problem you're fixing?

koriyoshi2041:
The regression is already included in this PR through the shared call-hierarchy helper in `tests/cpp/utils/ServerHarness.cpp`. Every existing cone case now checks that each returned item has `range == fromRanges[0]`, `selectionRange == fromRanges[0]`, and that the result count exactly matches the expected references. Without the source change, those assertions fail because the item ranges remain `{0,0}-{0,0}` and duplicate leaves make `result->size()` exceed the expected set.

For the issue example, the relevant JSON shape is:

```json
{
  "from": {
    "name": "test.the_sub_2.a",
    "uri": "file:///.../cone_test.sv",
    "range": { "start": { "line": "<cursor line>", "character": "<cursor column>" }, "end": { "line": "<token end line>", "character": "<token end column>" } },
    "selectionRange": { "start": { "line": "<cursor line>", "character": "<cursor column>" }, "end": { "line": "<token end line>", "character": "<token end column>" } }
  },
  "fromRanges": [{ "start": { "line": "<cursor line>", "character": "<cursor column>" }, "end": { "line": "<token end line>", "character": "<token end column>" } }]
}
```

The exact coordinates are derived from the existing `cone_test.sv` cursor, so the test does not hard-code brittle line numbers. Hosted CI is green across the full matrix at `ad752df`. If you would prefer this as a named standalone section in `ConeTests.cpp` rather than shared-helper coverage, I can split it out.

toddstrader:
We've been having some discussion about whether this LSP interface is the correct model or not.  I patched this and tried the vanilla neovim API calls and drivers worked the same, but loads only gave me a single blank entry to choose from, as opposed to before this change where it gave me the loads.

Actually, I'm now noticing your comment in `getCallHierarchyOutgoingCalls()` which is why this no longer works.  Well, the hard coded empty vector, not the comment.

I think we can go one of two ways here.  Either, make sure the base neovim API continues to work while also allowing `trouble.nvim` to work.  To do that we'll first need tests for the neovim API in `plugin_spec.lua`.  I'm not sure why the neovim API navigates to the correct document with the current implementation while `trouble.nvim` does not.  One would need to investigate.

Or alternatively, we could give up on mapping this to call hierarchy and roll our own interface.  This is already being discussed in #409 , but what is still missing is the client side.  We'd need neovim and vscode, I'm guessing you're more interested in the former.

Either way, we should lock this down with some lua tests.

koriyoshi2041:
Addressed this in `8f10988` without disabling outgoing loads.

The issue was that `CallHierarchyOutgoingCall.fromRanges` must refer to ranges in the caller (`params.item.uri`), but we were returning the target symbol's range there. Vanilla Neovim navigates through `to`, which explains why the old response worked there while clients that render `fromRanges` against the caller document showed the wrong content.

This revision keeps the target URI/range in `to` and uses the queried item's selection range for `fromRanges`. It adds both a C++ contract assertion and a `plugin_spec.lua` regression covering an outgoing `foo.a -> foo.b` load. Locally, the focused cone test passes all 120 assertions and pre-commit passes; I don't have `nvim`/`luarocks` available in this environment, so the hosted Neovim job is the remaining runtime check.

koriyoshi2041:
One additional wire-level check found that the prepared hierarchy item itself still carried default `(0,0)` ranges, so merely reusing `item.selectionRange` was protocol-valid but not useful.

`680f41e` now populates the prepared item's URI/range from the resolved symbol declaration and locks that down in both C++ and Lua. I also exercised the compiled server over actual JSON-RPC (initialize -> didOpen -> setTopLevel -> prepare -> outgoing): `foo.a` prepares with a non-empty declaration range, `foo.b` remains the `to` target, and `fromRanges` round-trips exactly to the caller item's selection range.

The new hosted workflows are again awaiting external-contributor approval; there is no failing job to diagnose yet.

AndrewNolte:
Will be merging https://github.com/hudson-trading/slang-server/pull/473

PR #455, opened 2026-08-14
Fix call hierarchy item source ranges

namtran1812:
## Summary

Fixes the source ranges reported on call hierarchy items for driver/incoming-call results.

Previously, `CallHierarchyItem.range` and `selectionRange` were left at their default `(0, 0)` values even though the correct source location was already available in `fromRanges`.

This caused clients that rely on the hierarchy item's own range to display or navigate to the wrong location.

## Changes

- Populate `CallHierarchyItem.range` with the cone leaf source range
- Populate `CallHierarchyItem.selectionRange` with the same source range
- Reuse the converted LSP range for `fromRanges`
- Add regression coverage verifying that `range` and `selectionRange` match the actual incoming-call source location

## Testing

```text
server_unittests "Cone Tracing"
All tests passed
```

Also verified the full test suite locally.

## Issue

Partial fix for #423.

I also investigated the duplicate cone-leaf behavior described in the issue. I tested both the existing multi-bit cone fixture and a minimized 4-bit continuous-assignment reproduction, but could not reproduce duplicate incoming hierarchy entries on current `main`, so this PR intentionally does not add deduplication logic without a failing regression.

### DIFF
--- a/include/ast/ServerCompilation.h
+++ b/include/ast/ServerCompilation.h
@@ -83,14 +83,22 @@ public:
             if (range.start().valid()) {
                 auto fullPath = std::filesystem::absolute(
                     m_sourceManager.getFileName(range.start()));
+                auto lspRange = toRange(range, m_sourceManager);
+
                 // only different by to / from field name . . . sigh
                 if constexpr (std::is_same_v<lsp::CallHierarchyIncomingCallsParams, P>) {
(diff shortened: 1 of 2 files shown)


AndrewNolte:
Will be merging https://github.com/hudson-trading/slang-server/pull/473

Issue #428, opened 2026-07-23
Diagnostics from full-build compilation get overwritten by stale shallow re-analysis of already-open documents on window reload


Praflower:
### Description
On a cold VS Code window reload (with several editor tabs already open/restored), the initial full-build
compilation runs first and correctly publishes diagnostics for all files in the build. However, shortly
after, the server re-analyzes each already-open document via `textDocument/didOpen` (shallow compilation),
and republishes diagnostics for those specific URIs — overwriting the correct full-build diagnostics with
the (often empty/incomplete) shallow-compilation results.

As a result, files that happen to be open in editor tabs at reload time show no (or incorrect) diagnostics,
while files not open in any tab correctly retain the full-build diagnostics.

Manually re-running "Select Build File" after the window has fully loaded works around the issue, because
it forces the full-build diagnostics to be published *after* the didOpen shallow passes, so they are the
last ones to win.

### Steps to reproduce
1. Open several `.sv` files that are part of the configured `build` (.f file), e.g. instantiated submodules.
2. Reload the VS Code window (so these tabs are restored on startup).
3. Observe the server log: the full build's diagnostics (`Driver analysis found N diagnostics`) are
   published first, followed by `textDocument/didOpen` shallow analyses for each restored tab.
4. Diagnostics in the editor for the restored tabs are missing/incorrect, even though the log shows the
   full build found real issues for those files.
5. Run "slang: Select Build File" and pick the same build file again — diagnostics now show correctly.

### Additional oddity observed
During the didOpen shallow re-analysis of two files that use a macro-only header (`\`include`d directly),
the log shows:

ERROR: Token '`SOME_MACRO': Directive overlaps with previous token '' : Identifier

This appears to be a token/position caching issue when the same file is parsed twice (once as part of the
full-build compilation, once again during the shallow per-document analysis), with different macro states.

### Expected behavior
Diagnostics published from the full-build compilation should not be silently overwritten by a later,
narrower shallow-compilation pass for documents that are already covered by the active build — or the
server should re-run/re-publish the build's diagnostics after all restored documents have been opened.

### Environment
- slang-server version: 0.2.9 (extension `hudson-river-trading.vscode-slang` 0.2.17)
- OS: Linux (remote SSH / VS Code Remote)
- Config: server.json with a `build` field pointing at a `.f` file (large multi-hundred-file RTL project)

AndrewNolte:
Are you using `--single-unit`?

PR #457, opened 2026-08-15
Fix unchanged file reloads overwriting compilation diagnostics

pravinireri:
## Summary

* Reloading an open document from disk could re-publish shallow diagnostics even when its contents had not changed, replacing compilation diagnostics for that file.
* `SlangDoc::reloadBuffer` now skips the reload when the on-disk contents already match the current buffer.

## Why

#311 prevents matching `didOpen` contents from replacing compilation diagnostics with shallow diagnostics. The reload path could still cause the same downgrade because `reloadBuffer()` replaced the buffer unconditionally before `updateDoc(CHANGE)`. Genuine external changes continue through the existing reload path.

## Tests

* Added `UnchangedWatchedFileDoesNotOverwriteCompilationDiags`
* On unmodified main, the regression test fails with 1 compilation diagnostic replaced by 18 shallow diagnostics.
* Relevant diagnostic, watcher, and external-file-change tests pass.

Refs #428

### DIFF
--- a/include/document/SlangDoc.h
+++ b/include/document/SlangDoc.h
@@ -129,7 +129,7 @@ public:
     void onChange(const std::vector<lsp::TextDocumentContentChangeEvent>& contentChanges);
 
     /// @brief Re-read the buffer from disk (used for external file changes)
-    /// @return true if successful, false if the read failed
+    /// @return true if the contents changed, false if unchanged or the read failed
     bool reloadBuffer();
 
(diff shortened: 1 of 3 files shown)


AndrewNolte:
This doesn't address the issue  - reloadBuffer() is called for workspace updates- like rebasing your branch. The issue is about `textDocument/didOpen` overwriting the diags. I have this issue addressed on a branch that I'll be submitting soon.

PR #475, opened 2026-08-19
LSP Logging: Log JSON-RPC request latency

AndrewNolte:


AndrewNolte:
Looks like this broke windows builds - and should've ran the full CI. Will revert and fix.

PR #478, opened 2026-08-20
Build: Give SyntaxTarget an explicit ctor for GCC 13

### LINKED ISSUE #407: GCC 13 Ubuntu 24.04 WSL Build Error

Image

shubhxho:
Fixes #407

GCC 13 rejects the default member initializer on `SyntaxTarget::macroUsageRange` because `SyntaxTarget` is a nested class used as a `std::variant` alternative of the enclosing type. The default-ctor check needs that initializer before the enclosing class is complete.

Move the `SourceRange::NoLocation` default onto the constructor instead. Clang / GCC 14 already built this; this just unblocks GCC 13.

### DIFF
--- a/include/document/DefinitionInfo.h
+++ b/include/document/DefinitionInfo.h
@@ -41,7 +41,14 @@ struct DefinitionInfo {
         // found.
         slang::parsing::Token nameToken;
         // Optional original source range; exists if it's behind a macro expansion.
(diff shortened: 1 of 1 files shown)


shubhxho:
Superseded by #480, which has this commit plus the #428 diagnostic fix.

PR #480, opened 2026-08-20
Build/Diagnostics: GCC 13 ctor + keep design diags on didOpen

### LINKED ISSUE #407: GCC 13 Ubuntu 24.04 WSL Build Error

### LINKED ISSUE #428: Diagnostics from full-build compilation get overwritten by stale shallow re-analysis of already-open documents on window reload

shubhxho:
Fixes #407
Fixes #428

Two small, independent fixes:

- **#407** — `SyntaxTarget` is a nested class used in a `std::variant` of the enclosing type. GCC 13 rejects the in-class default for `macroUsageRange`. Move that default onto the constructor.
- **#428** — Restored tabs on window reload ran a shallow `didOpen` pass and overwrote full-build diagnostics. The skip only fired on an exact buffer match (including SourceManager's trailing NUL). Skip the shallow pass whenever the URI is already in the compilation.

`server_unittests`: 241/241 passed.

### DIFF
--- a/include/document/DefinitionInfo.h
+++ b/include/document/DefinitionInfo.h
@@ -41,7 +41,14 @@ struct DefinitionInfo {
         // found.
         slang::parsing::Token nameToken;
         // Optional original source range; exists if it's behind a macro expansion.
-        slang::SourceRange macroUsageRange = slang::SourceRange::NoLocation;
+        slang::SourceRange macroUsageRange;
+
+        // Don't default-init macroUsageRange in-class: GCC 13 can't see that initializer when
+        // SyntaxTarget is used in a std::variant of the enclosing type.
+        SyntaxTarget(const slang::syntax::SyntaxNode* node, slang::parsing::Token nameToken,
+                     slang::SourceRange macroUsageRange = slang::SourceRange::NoLocation) :
+            node(node), nameToken(nameToken), macroUsageRange(macroUsageRange) {}
+        SyntaxTarget() : SyntaxTarget(nullptr, {}, slang::SourceRange::NoLocation) {}
 
         bool operator==(const SyntaxTarget& other) const {
             return node == other.node && nameToken.location() == other.nameToken.location() &&
--- a/src/ServerDriver.cpp
+++ b/src/ServerDriver.cpp
@@ -181,15 +181,19 @@ std::unique_ptr<ServerDriver> ServerDriver::create(Indexer& indexer, SlangLspCli
             // Only copy if the URI isn't already in the new driver's docs
             auto newDocit = newDriver->docs.find(uri);
             if (newDocit == newDriver->docs.end()) {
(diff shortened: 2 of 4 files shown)


AndrewNolte:
Please split these into two PRs, and add gcc-13 to CI for that fix. 

For the design diags one that test is not what we want- the design diagnostics will not map to an actual source range in the document. Always skipping is not the correct solution. I will be addressing this on a focused-instance branch.

shubhxho:
Thanks for the review — agreed on both points.

- Re-opened the GCC 13 fix on its own (with an explicit `g++-13` CI job): #484
- Dropping the #428 approach here; leaving that for your focused-instance branch.

PR #488, opened 2026-08-23
skip block clauses tokens in `textDocument/references`

evanwporter:
Consider I have a module:

```
module MODULE_NAME;

endmodule : MODULE_NAME // <-- I really like adding this but I don't need it 
                        //       popping up in my references since it clogs 
                        //       up the list.
```

This is a personal preference (though I think its a sensible preference), so I'm okay with adding a entry in the config file to opt into this.

Also I made it so that it doesn't skip the clauses for begin end blocks, though I'm not sure why you'd ever want to look at the references for these.

### DIFF
--- a/include/ServerDriver.h
+++ b/include/ServerDriver.h
@@ -129,7 +129,8 @@ public:
     std::optional<std::vector<lsp::Location>> getDocReferences(const URI& uri,
                                                                const lsp::Position& position,
                                                                bool includeDeclaration,
-                                                               const lsp::RequestContext& ctx = {});
+                                                               const lsp::RequestContext& ctx = {},
+                                                               bool excludeEndBlockClauses = false);
 
     /// @brief Renames a symbol in a document
     /// @param uri The URI of the document
--- a/include/document/ShallowAnalysis.h
+++ b/include/document/ShallowAnalysis.h
@@ -168,8 +168,8 @@ public:
     /// @param targetLocation The source location of the target symbol
     /// @param targetName The name of the symbol to match
     void addLocalReferences(std::vector<lsp::Location>& references,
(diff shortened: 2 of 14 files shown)


AndrewNolte:
The rename route also uses this method- we don't want to skip it when renaming.

evanwporter:
> The rename route also uses this method- we don't want to skip it when renaming.

nice, didn't think about that. That would be unfortunate if this pr had gone through.

evanwporter:
@AndrewNolte ready for review

AndrewNolte:
I thought this was going to be behind a config? 

This is also just kind of a lot for such a small gain.

What if instead we added inlay hints for these? Then you also wont have to type them. 

Also this recent commit should help with indicating what scope you're in- https://github.com/hudson-trading/slang-server/pull/515. Not sure if your neovim setup utilizes it though.

evanwporter:
I use references a lot to navigate to where a module is used (quicker than
searching)

And the test files balloon up the pr but really it’s just a conditional
check.

I guess we can add a config but it seems kind of niche. And this seems in
my opinion like a very sensible change.

Inlay hints would be great here except in neovim I’ve been having issues
with them. Could be an issue with my config.

AndrewNolte:
It just adds a single reference though? And it's a reference in the same file as the module?

evanwporter:
@AndrewNolte can we reopen this; i find this valuable since in neovim its a little different than vscode:

image

All the references are globbed together in one list. In vscode you kinda click through the lists, so its not big deal

evanwporter:
Also FYI one side effect of switching too neovim is becoming much more willing to spend 6 hours to shave a single keystroke off your workflow

PR #492, opened 2026-08-23
Hovers: Show drivers propagated through ports

evanwporter:
image

This PR shows when ports are driven indirectly:

```
assign MODULE_INSTANCE_NAME.PORT_NAME =  LOGIC
```

stacks onto #489

### DIFF
--- a/src/document/DefinitionInfo.cpp
+++ b/src/document/DefinitionInfo.cpp
@@ -243,6 +243,11 @@ bool isPortDriver(const analysis::ValueDriver& driver) {
            driver.flags.has(analysis::DriverFlags::ViaIndirectPort);
 }
 
+bool isLogicThroughPort(const analysis::ValueDriver& driver) {
+    return driver.flags.has(analysis::DriverFlags::ViaIndirectPort) &&
+           !driver.isUnidirectionalPort();
+}
+
 std::string getDriverInstanceDescription(const DriverGroup& group) {
     const ast::InstanceSymbol* instance = nullptr;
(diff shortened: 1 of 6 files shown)


evanwporter:
@AndrewNolte I'm happy to change the test names but I was having a spasm trying to read this:

`HoverNestedGeneratedSymbolsDeduplicatesEquivalentElaborations`

So I set it to something nice and easy to read, but now I finished the patch I can change the names.

AndrewNolte:
Do you mind showing a more useful example? Bc you're hovering over the driver in that screenshot

evanwporter:
> Do you mind showing a more useful example? Bc you're hovering over the driver in that screenshot

yeah I'll take a nicer screenshot later this week

src/document/DefinitionInfo.cpp
```diff
@@ -243,6 +243,11 @@ bool isPortDriver(const analysis::ValueDriver& driver) {
            driver.flags.has(analysis::DriverFlags::ViaIndirectPort);
 }
 
+bool isLogicThroughPort(const analysis::ValueDriver& driver) {
+    return driver.flags.has(analysis::DriverFlags::ViaIndirectPort) &&
+           !driver.isUnidirectionalPort();
```
AndrewNolte, line 248: Are `ViaIndirectPort and isUnidirectionalPort` even possible? I don't think so

AndrewNolte:
A more accurate title would be mentioning interface ports. Same thing applies to isLogicThroughPort. reading `!isLogicThroughPort(*driver) && isPortDriver(*driver)` made no sense

Issue #355, opened 2026-05-27
[VScode] Add top level panel to sidebar


evanwporter:
A top level panel would be nice to add to the sidebar.

<img width="1128" height="478" alt="Image" src="https://github.com/user-attachments/assets/a5f44bd1-06cc-42a8-b1b1-adc0233bd352" />

In most cases the top level is a testbench that wraps the DUT. So, it would be nice to select between different top levels. Selecting a top level would change the hierarchy panel to display everything under the selected top level.

The other thing is that waveforms could be made to be associated with a top level

```
top-level
 ├── Waveform 1
 └── Waveform 2
```

These waveforms would be saved (see #206). (Note that multiple waveforms for a top level is pretty specific for my case as I use verilator to run different tests for a single top-level but ymmv).

Wave forms that have been added will have buttons to open the waveform in vaporview.

```
top-level
 ├── Waveform 1 <OPEN BUTTON>
 └── Waveform 2 <OPEN BUTTON>
```

I foresee this as a three-part PR. The first adds the top-level panel and the switching mechanism, the second part adds the waveform, and a third future part persistence across sessions.

PR #504, opened 2026-08-25
Persist active build/top-level selection across sessions

### LINKED ISSUE #206: [FEAT] top module setting
I've been trying to use the sidebar more, as I think that'll make me more productive, but one thing I find annoying is that Slang-server doesn't hold state in vscode; ie: you close the window and reopen it, and you have to reset your top-module. So, I was thinking a top module setting or something to go in the flag file would be cool.

I'm happy to make these changes myself, but I wanted to discuss first because there's a lot of different ways, we could solve this problem.

Abhayindia:
Fixes #206. Right now the active build file/command/top-level is only held in memory on `ProjectComponent`, so it's lost every time the window reloads. This persists it to `.slang/local/memento.json` in the workspace whenever it changes, and restores it on activation before the tree view does anything - the same design you and Todd landed on in the issue thread (plain JSON, client-agnostic, not the VS Code memento API).

Scoped to just the persistence piece so it doesn't overlap with #355 - that issue's own writeup already defers its "part 3: persistence" back here.

There's no VS Code extension host available in this environment to run a full integration test, so I verified the save/load/clear round-trip directly against a real temp workspace directory, including that a missing or corrupted memento file falls back to "no prior selection" instead of erroring:

```
$ ts-node verify_memento.ts
ok - no memento initially
ok - filelist round-trip
ok - topfile round-trip
ok - clearing removes memento
ok - corrupt memento treated as absent
All memento round-trip checks passed.
```

Also clean on the existing type-check and lint:

```
$ tsc --noEmit
OK: no type errors

$ eslint src/sidebar/ProjectComponent.ts src/sidebar/ProjectMemento.ts
OK: no lint errors
```


AndrewNolte:
This should be done server side if at all. That was the reasoning behind not using the vscode api

Abhayindia:
makes sense - todd said the same in #206. want me to move it into ServerDriver (server owns .slang/local/memento.json, pushes the restored build/top selection to the client on init), or is this not worth pursuing right now?

Abhayindia:
@AndrewNolte this one's gone stale - main has moved hierarchy opens server side since (#526) and ProjectComponent.ts got rewritten around it, so the client-side memento here doesn't really fit any more and the branch conflicts.
happy to just close it, or to redo it in ServerDriver the way you and todd suggested - which would you rather?

Abhayindia:
rebased and reworked against the refactored ProjectComponent - the getter/setter and the restore both still fit main's structure, and the restore now runs at the end of `activate()` rather than where it used to sit. `tsc --noEmit` is clean on src.
your server-side point still stands though, so this is only the conflict cleared, not an argument for the approach - happy to move it into ServerDriver or close it, whichever you'd rather.

AndrewNolte:
I thought we were moving this into the server? So cpp side? And using reflect cpp to define the schema.

Let's also just call it state.json

Abhayindia:
sounds good - server side in ServerDriver, `rfl` schema, `state.json`. one thing i'd rather settle before writing it: `.slang/local/` is the natural home next to the existing local `server.json`, but the watcher at `SlangServer.cpp:313` globs `.slang/**/*.json` and `onWorkspaceDidChangeWatchedFiles` flags `needsReload` for any `.json`, so persisting on every build/top switch would fire `loadConfig()` + `setBuildFile()` each time.
do you want `state.json` outside that glob, or the handler taught to skip it? and per-workspace only, or with a `~/.slang/` fallback like `server.json` has?

AndrewNolte:
Yeah let's narrow that glob trigger to just the files it needs

Abhayindia:
Put the glob narrowing up separately as #551 so it can land ahead of the `ServerDriver` work: `loadConfig()` only reads `.slang/server.json` and `.slang/local/server.json`, so both the watcher and the blanket `.json` check in `onWorkspaceDidChangeWatchedFiles` now match just those two.
Fair warning that I could not compile it here (needs CMake 3.28 and g++-14, this box has 3.22 and g++ 12), so CI is the real check; I did compile and table-test the predicate on its own. Say the word if you would rather it came bundled with the state.json change instead.

Issue #301, opened 2026-04-10
[feat] quick fix imports


evanwporter:
Both the clangd and typescript lsp have the option to quick fix include stuff. A similar feature in slang-server would be great. The feature would probably work like typescript when you want to import a package, and it would look like clangd when you want to include a SV header file.

I'll add images later with more context.

This can be done using code actions.

Ill also see if I can dig up the relevant clangd commits.

Some of the work has already been done in #299.

AndrewNolte:
Sounds great! Yeah I think we'd want to do this for unknown macros + adding include files to start out. I'm not sure if this makes sense to do this for packages. Imo `import SomePkg::*` is not a great practice, and I don't think we should start indexing packages just to support this feature.

evanwporter:
I was thinking that it does import SomePkg::{symbol1, symbol2} like in
typescript. Rather than importing everything. It could also qualify the
symbol like SomePkg::symbol1. A drop down could allow choosing between the
two.

evanwporter:
> Sounds great! Yeah I think we'd want to do this for unknown macros + adding include files to start out. I'm not sure if this makes sense to do this for packages. Imo `import SomePkg::*` is not a great practice, and I don't think we should start indexing packages just to support this feature.

@AndrewNolte can we revisit this. In my  workflow (which may be unconventional :/ ), I run into the variable not found error 15 times a day because I need to import the package.

Then I need to recall which package has the type. Then I usually do import pkg::*.

But what I’d find useful is code action like in typescript + other lsps, to auto import the correct variable via a code action. Auto import, auto qualify.

I feel like this could be easily done by storing a dict like so:

dict[ type ptr, list[pkg ptr] ]

The once a type is typed out we have a list of pkgs that hold said list type.

Might require a change to vendored slang to get it working efficiently.

Abhayindia:
Looking at picking up the narrower half of this - unknown macro -> quick fix to add the `` `include `` for it, leaving the package auto-import part out per your comment above (I'd rather not wade into that debate).

[...]

PR #508, opened 2026-08-27
Add code action to `include an undefined macro known to the index

Abhayindia:
Implements the unknown-macro half of #301, per the scope agreed with @AndrewNolte in that thread (undefined macro -> add a `include for it; the package auto-import part is left out).

When the cursor is on a `` `MACRO `` whose definition isn't visible in the current compilation, and the workspace index knows a file that defines it (`Indexer::getFilesForMacro`, already populated for macro go-to-definition), a `Add \`include "..."` quick fix is offered - one per candidate file. The new `include is placed after any existing ones at the top of the file, using a path relative to the current file, falling back to the bare filename when that would climb out of the tree. It's suppressed when the macro is actually defined locally, when the file is already included, or when the index doesn't know it.

An undefined macro usage leaves no token in the shallow analysis' token index, so the usage node is looked up by cursor position through `collectedHints` instead of the usual `getWordTokenAt` path.

Tests are in `CodeActionTests.cpp` (offered case, plus the three suppression cases). Verified locally against the real build:

```
$ build/bin/server_unittests "*CodeAction*"
All tests passed (27 assertions in 11 test cases)
```

The full `server_unittests` binary links and runs here only with a couple of local-only MinGW linker workarounds for a pre-existing Windows build issue in `WcpClient` (same one noted in #499); the two unrelated failures left in the full suite (`URITests`, `DriverTests`) are Windows path-handling and also present on a clean checkout.


AndrewNolte:
In this impl, later defines in the file would suppress valid fixes.

src/codeactions/IncludeForMacro.cpp
```diff
@@ -0,0 +1,122 @@
...
+        if (macroPath == thisFile || fs::equivalent(macroPath, thisFile, ec))
+            continue;
+        if (std::ranges::any_of(alreadyIncluded, [&](const fs::path& p) {
+                std::error_code e;
+                return p == macroPath || fs::equivalent(p, macroPath, e);
+            }))
+            continue;
+
+        // Prefer a path relative to the current file; fall back to the bare
+        // filename when the relative path would climb out of the tree.
```
AndrewNolte, line 100: This can result in adding an invalid include if the folder isn't an incdir? "The tree" here appears to just be the file's parent folder, so no relative include would ever get produced?

Abhayindia:
on the later-define: right, `analysis.macros` is the end-of-preprocess set so a define anywhere in the file suppresses it. dropping that check and keying only off `macroUsageDefinitions` (which only holds refs that actually expanded) fixes it - a define that comes after the use, or after an `undef, stays unresolved. will add a test for that case.

on the include path: agreed the relative logic mostly collapses to the bare filename and doesn't know the incdirs. do you want it resolved against the project's include dirs (is that list reachable from the code-action ctx?), or just always emit the bare filename and leave incdir setup to the user?

Abhayindia:
pushed the later-define fix - dropped the `analysis.macros` check so it keys only off `macroUsageDefinitions`, plus a `WhenDefinedLater` test covering a define that appears after the use. couldn't build locally (the pinned ctre submodule commit isn't fetchable any more, same on main), so i'm leaning on CI here.
still unsure on your incdir point - should the include resolve against the incdirs the index knows, or stay a bare filename? happy to go either way.

AndrewNolte:
This still:
- Adds the include after the last include, which may be after the macro usage
- Fails to take into account include paths that were added - The added code should never add a new diagnostic for a missing include path

Please at least try this out on an actual project

AndrewNolte:
I'll implement this in an inc path indexing / UVM pkg pattern support PR

Abhayindia:
understood, and thanks for the detailed review - both points are fair. i should have tried it against a real project before pushing rather than reasoning it through, and the include-path one especially could have made things worse.
good luck with the inc path indexing PR, happy to test that one against a real build if it would help.

Issue #509, opened 2026-08-27
LSP positions are off on lines containing non-ASCII characters (byte vs UTF-16 column)


Abhayindia:
Positions get misaligned on any line that contains non-ASCII characters. Hover ranges, go-to-definition highlights, diagnostic squiggles, document highlights and rename edits all land a few columns to the right of where they should, and the shift grows with each multibyte character earlier on the line.

Minimal repro:

```systemverilog
module m;
  //°C threshold
  logic clk;   // hover over clk highlights "lk;" instead
endmodule
```

The `°` is two bytes in UTF-8, so everything after it on that line (and the ranges the server reports for tokens on later lines that share a source line with a comment like this) is off by one. A comment with a few CJK characters, or a string literal like `$display("µs")`, pushes it further.

The cause is in `toPosition` (`src/util/Converters.cpp`): it uses slang's `SourceManager::getColumnNumber`, which is a raw byte offset into the line, directly as the LSP `character` value. LSP counts `character` in UTF-16 code units unless the server negotiates otherwise, and `getInitialize` doesn't set `positionEncoding`, so clients assume UTF-16. `toSourceLocation` has the same problem in reverse, so requests made with the cursor after a multibyte character on a line resolve to the wrong token too.

The usual fix is to advertise `positionEncoding: "utf-8"` when the client lists it in `general.positionEncodings` (most modern clients do), and otherwise translate between the byte column and the UTF-16 column using the line text. slang already exposes the line contents so the conversion doesn't need a new data source.

Abhayindia:
I'd like to take this one. Would you want it to negotiate `positionEncoding: utf-8` when the client supports it and only convert byte<->UTF-16 as a fallback, or just always convert and keep the wire protocol on UTF-16?

AndrewNolte:
Let's advertise utf 8, and start with just fixing /didChange updates with non-ascii chars. I don't think it'll be worth fixing all the other `getColumnNumber ` calls since non-ascii chars can only show up in comments, not code.

Abhayindia:
#511 does that now - advertises utf-8 and fixes the didChange offsets.

PR #511, opened 2026-08-28
negotiate utf-8 position encoding, convert didChange offsets on non-ascii lines

Abhayindia:
Reworked per your note on #509. Instead of converting in `toPosition`/`toSourceLocation`, the server now advertises `positionEncoding: utf-8` when the client offers it in `general.positionEncodings`. slang already works in byte columns, so on that path there's nothing to convert and the other `getColumnNumber` call sites are left alone.

didChange still needs handling: `getOffsets` was adding `Position.character` straight onto a byte line offset, so an incremental edit after a non-ascii character on the same line landed at the wrong spot and corrupted the buffer. It now adds `character` directly when utf-8 was negotiated, and walks the line to convert from utf-16 code units otherwise.

`utf16ColumnToByte` lives in `Converters.cpp` with a test in `ConverterTests.cpp` (two-byte char + a non-BMP one). `server_unittests` doesn't link under MinGW here, pre-existing per #499, so I built and ran the converter tests against the real `libsvlang`:

```
All tests passed (31 assertions in 3 test cases)
```

The utf-16 branch in `getOffsets` is a few lines - happy to drop it and rely purely on negotiation if you'd rather.

Addresses #509.


Abhayindia:
@AndrewNolte when you get a chance, could you take a look? This is the #509 position-encoding bug. #499, #504 and #508 are also still open from the last couple weeks if you're doing a review pass.

AndrewNolte:
Left a comment on the issue- lets start with just handling didChange updates for variable length utf8 chars properly, and advertising that in the capabilities

Abhayindia:
@AndrewNolte this is reworked to match that - it advertises `positionEncoding: utf-8` when the client offers it, and fixes the `onChange` offset math so variable-length utf8 didChange edits apply at the right byte. ready for another look when you have a minute.

Abhayindia:
@AndrewNolte bump on this when you get a chance.

PR #529, opened 2026-09-02
Implement Active Instances

AndrewNolte:
- Hierarchy View
  - Show data by default now
  - Clicking sets the active instances for module up to the selection
- Modules View
  - Shows only the opened module's instances
  - Clicking now takes you to active instance; The 'file' button now takes you to the instantiation
- Adds codelenses for active instance to select, along with goto instantiation lens
- Switch to server-supported search that now includes arbitrary symbols, not just hierarchy
- Re-target symbols onto saved AST, showing params values, iface defs, etc from focused instance.
  - Show active parameter values as inlay hints
  - Gotodefs on iface ports go up the iface port chain to the instantiation
  - More work to be done here to make routes like goto def / refs apply for just the selected design

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,6 +291,7 @@ add_library(
   slang_server_obj_lib OBJECT
   ${CMAKE_CURRENT_BINARY_DIR}/VersionInfo.cpp
   src/SystemTaskDocs.cpp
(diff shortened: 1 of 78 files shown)


AndrewNolte:
Re-stacked at https://github.com/hudson-trading/slang-server/pull/538; search feature was split into separate PR

